PR #9, opened 2024-11-11
feat: Add environment variable validation and secure key management

robertisaacs:
• Add env validation with Zod schema
• Implement secure key auto-generation
• Add predev/prestart validation hooks
• Create env config processing system
• Move server config to dedicated files


robertisaacs:
Right now the startup of the application can fail and not give very helpful information on why it isn't starting correctly or what is missing. This change uses zod to validate the .env file on startup, fills in random values for missing cryptographic tokens, etc. It gives a nice indication of which fields are missing and cannot be defaulted.

PR #14, opened 2024-11-11
Add shared document types and make document type field nullable

RobertAtNineMinds:
• Adds shared_document_types table
• Makes document type_id nullable
• Seeds common MIME types
• Updates document preview generation
• Adds type detection for uploads


RobertAtNineMinds:
In order to avoid duplicating document types (mime types, essentially) for each tenant, we need a shared document type table where we don't require the same RLS policies and tenant fields. Then, all of the corresponding logic needs to be updated to accommodate the new table structure.

Issue #27, opened 2024-11-14
Time entry / add new work item / a long list of work items is not navigable and hard to read


rjanicek:
Can not scroll up or down, also can not read the entire work item title if it contains long words (like URL's) which are not wrapped.

![image](https://github.com/user-attachments/assets/fce7e409-1fc3-4262-86c5-f3e228b37f90)

RobertAtNineMinds:
fixed by #62

PR #62, opened 2024-11-22
Fix scrolling behavior in work item picker dialog

RobertAtNineMinds:
• Add fixed height and scrollable container
• Move search box to sticky header
• Improve overflow handling
• Clean up flexbox layout structure


RobertAtNineMinds:
Fixes #27

PR #501, opened 2025-05-21
Add version tag triggers and dynamic Docker tagging to publish workflow

NatalliaBukhtsik:
- Added push/create triggers for version tags (v*.*.*)
- Modified publish job condition to handle multiple event types
- Implemented dynamic Docker tag determination based on event type
- Consolidated tag output into single step

Impact: More flexible publishing options for both CI and manual version releases.

Like a mad scientist's experiment, these tags shall now multiply uncontrollably! May the version gods have mercy on our CI pipeline! 🧪🔖🤪


NatalliaBukhtsik:
The code gremlins have been fed after midnight - what could possibly go wrong? 🎲👹

PR #917, opened 2025-08-02
Added priority color display to ticket details

NatalliaBukhtsik:
- Extended TicketWithDetails interface to include priority_color
- Updated ticket details UI to show colored priority indicator
- Modified client-tickets actions to include priority color in queries
- Added styling for priority color display in badge component

These changes affect both frontend display and backend data fetching, ensuring consistent priority color representation throughout the client portal.

"Now the tickets shall wear their priorities like the Cheshire Cat wears its grin - with colorful flair and disappearing logic!" 🎩🐇🌈

### DIFF
--- a/server/src/components/client-portal/tickets/TicketDetails.tsx
+++ b/server/src/components/client-portal/tickets/TicketDetails.tsx
@@ -37,6 +37,7 @@ interface TicketDetailsProps {
 interface TicketWithDetails extends ITicket {
   status_name?: string;
   priority_name?: string;
+  priority_color?: string;
   conversations?: IComment[];
   documents?: IDocument[];
   userMap?: Record<string, { first_name: string; last_name: string; user_id: string; email?: string; user_type: string; avatarUrl: string | null }>;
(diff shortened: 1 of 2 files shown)


robertisaacs:
## 🤖 AI Code Review

Great quality-of-life PR – showing the ticket’s priority colour makes the list and the details view much more scannable, and the implementation touch-points are small and well isolated (query update + UI tweak). 👍

Critical / must-fix
1. Inline style with un-sanitised DB value  
   priority_color is rendered directly into a style attribute. If a tenant manages to store something like `url("javascript:alert(1)")` they could execute script in the client portal.  
   • Please whitelist/validate the colour on the server (e.g. hex regex) or map priority→colour on the client instead of trusting raw DB input.  
   • Using a Tailwind class (e.g. bg-\[${color}\]) via the clsx helper would also avoid inline styles and keep CSP tighter.

2. Hard-coded fallback colour  
   Rule 4 forbids hard-coded values in UI. Replace `'#6B7280'` with a value from the theme (e.g. `text-gray-500`/`bg-gray-500`) or pull it from a constants file so theming stays centralised.

Questions / double-check
• Citus tenant filter – the modified queries still look correct, but can you confirm that getClientTicketDetails and getClientTickets include the usual `.where('t.tenant', tenantId)` earlier in the chain? I can’t see that part in the diff.  
• Is priorities a reference table? If it’s distributed you’ll need `on('t.tenant', '=', 'p.tenant')` to satisfy the “JOINs must include tenant” rule.  
• Do we need to extend the client-portal RBAC layer to expose `tickets.read` for the new field, or is that already covered?

Nice-to-have / suggestions
• Instead of nesting a div inside a span, consider switching the outer element to a flex container to keep correct semantics.  
• We already ship a Badge component in `@/components/ui`; swapping the manual styling for that would keep appearance consistent.  
• If you want a visual indicator in lists as well, you can reuse this snippet in the table row component for parity.

Overall this is a tidy enhancement – fix the inline-style safety/branding points and we’re good to go. Thanks for the improvement!

---

PR #918, opened 2025-08-02
fix: remove documents PVC from dev deployment

robertisaacs:
The dev (sebastian-dev) Deployment was mounting the documents PersistentVolumeClaim, which is not multi-attach capable. This change removes the storage-volume PVC and mount from deployment-dev.yaml so the PVC is only bound to the main sebastian Deployment.

### DIFF
--- a/helm/templates/deployment-dev.yaml
+++ b/helm/templates/deployment-dev.yaml
@@ -52,16 +52,6 @@ spec:
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
       volumes:
-        {{- if .Values.server.persistence.enabled }}
-        - name: storage-volume
-          {{- if .Values.server.persistence.existingClaim }}
-          persistentVolumeClaim:
-            claimName: {{ .Values.server.persistence.existingClaim }}
-          {{- else }}
-          persistentVolumeClaim:
-            claimName: {{ include "sebastian.fullname" . }}-storage
-          {{- end }}
-        {{- end }}
         # Source code volume (emptyDir for fresh clone each pod start)
         - name: source-code
           emptyDir: {}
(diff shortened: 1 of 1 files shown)


robertisaacs:
## 🤖 AI Code Review

Nice quality-of-life tweak!  Dropping the PVC from the dev deployment will speed up pod start-up and avoids dangling volumes on every “helm upgrade”.  The change is narrowly scoped to helm/templates/deployment-dev.yaml and leaves prod charts untouched – good separation.

CRITICAL / MUST FIX
None blocking spotted.  The generated YAML still contains at least one item under
spec.template.spec.volumes (source-code) and spec.template.spec.containers[*].volumeMounts, so the manifest stays syntactically valid.

SUGGESTIONS & QUESTIONS
• Confusing values: If values-dev.yaml still has server.persistence.enabled=true (or the mountPath defined), Helm users may wonder why no volume shows up.  Consider setting server.persistence.enabled: false in the dev values file or adding a comment there that the setting is intentionally ignored by deployment-dev.yaml.  
• Dead code?  Now that deployment-dev no longer references .Values.server.persistence.*, you might remove that stanza (or move it behind a common “prod-only” block) to keep the chart tidy.
• Runtime check: Make sure no container process tries to write to the path previously mounted via .Values.server.persistence.mountPath; otherwise we’ll get “read-only file system” errors at runtime.  A quick grep in /app for that path might be worthwhile.
• Helm lint: After the change, run “helm template . | kubectl apply ‑-dry-run=client -f -” to double-check indentation – comments were re-indented manually and it’s easy to miss a space.

Otherwise everything looks consistent with our patterns: no RBAC, DB, API or UI code touched.

Thanks for the cleanup – this should make local dev a bit smoother!

---

PR #920, opened 2025-08-03
fix: remove verbose logging from secret retrieval and update import path

robertisaacs:
## Summary
- Remove verbose logging from `get_secret` function in entrypoint scripts
- Update logger import path in NextAuth options

## Changes
1. **server/entrypoint.sh** and **services/workflow-worker/entrypoint.sh**:
   - Removed log messages when using environment variables instead of Docker secrets
   - Removed log messages when neither secret file nor environment variable is found
   - These messages were creating unnecessary noise in logs

2. **server/src/app/api/auth/[...nextauth]/options.ts**:
   - Updated logger import from `@alga-psa/shared/dist/core/logger.js` to `@alga-psa/shared/core/logger.js`
   - This uses the direct import path instead of the built distribution path

## Test plan
- [ ] Verify services start correctly without the verbose logging
- [ ] Confirm secrets are still properly loaded from both Docker secrets and environment variables
- [ ] Verify NextAuth functionality works with updated import path

### DIFF
--- a/server/entrypoint.sh
+++ b/server/entrypoint.sh
@@ -16,10 +16,8 @@ get_secret() {
     if [ -f "$secret_path" ]; then
         cat "$secret_path"
     elif [ ! -z "${!env_var}" ]; then
-        log "Using $env_var environment variable instead of Docker secret"
         echo "${!env_var}"
     else
-        log "Neither secret file $secret_path nor $env_var environment variable found, using default value"
         echo "$default_value"
     fi
 }
--- a/server/src/app/api/auth/[...nextauth]/options.ts
+++ b/server/src/app/api/auth/[...nextauth]/options.ts
@@ -7,7 +7,7 @@ import { authenticateUser } from "server/src/lib/actions/auth";
 import { getKeycloakToken } from "server/src/utils/keycloak";
 import { decodeToken } from "server/src/utils/tokenizer";
 import User from "server/src/lib/models/user";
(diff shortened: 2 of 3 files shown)


robertisaacs:
## 🤖 AI Code Review

Nice tidy-up! 🎉  This PR silences the noisy “secret” messages in the container entrypoints and switches the NextAuth options file to the non-dist logger import, keeping the codebase a bit cleaner.

Critical issues  
None detected.  The changes are restricted to log statements in two shell scripts and a single import path; no RBAC calls, SQL, or UI code is touched, so the critical patterns remain unaffected.

Suggestions / questions  
• Shell scripts – consider leaving a debug-level branch (e.g. `if [[ "$LOG_LEVEL" == "debug" ]]; then log "..."; fi`) instead of a hard removal.  It can be handy when diagnosing misconfigured secrets in prod without having to re-add statements.  
• Logger import – can you confirm that `@alga-psa/shared/core/logger.js` is published in the package root of `@alga-psa/shared`?  Consumers that still run from the compiled package (e.g. background workers) might fail if `core/logger.js` is only present in `dist/`.  A quick `npm pack` test of the shared package would catch this.  
• Unused `log` function?  After dropping the two calls, the `log()` helper defined earlier in each script may now be unused.  If so, we could remove the helper entirely to avoid dead code.

All in all, very small, low-risk changes—thanks for trimming the noise! 🚀

---

PR #921, opened 2025-08-03
Fix vault agent secret injection and related improvements

robertisaacs:
This PR addresses several issues with vault agent secret injection:

- Remove unnecessary command override for vault agent in deployment template
- Trim whitespace in vault agent secret injection template and remove redundant args
- Update vault agent secret injection keys and remove unnecessary logging
- Remove template-style vault agent secret injection for alga_auth_key
- Include nextauth secret in temporal worker startup validation
- Remove fallback logic for missing enc keys
- Remove redundant fallback

These changes improve the reliability and security of secret injection in the vault agent setup.

### DIFF
--- a/ee/temporal-workflows/src/config/startupValidation.ts
+++ b/ee/temporal-workflows/src/config/startupValidation.ts
@@ -13,6 +13,7 @@ const logger = createLogger({
 const REQUIRED_CONFIGS = {
   // Authentication
   ALGA_AUTH_KEY: 'Shared authentication key for internal API calls',
+  NEXTAUTH_SECRET: 'Secret used by NextAuth for session management and password hashing',
   
   // Database
   DB_HOST: 'Database host (e.g., localhost or postgres)',
--- a/helm/templates/deployment-dev.yaml
+++ b/helm/templates/deployment-dev.yaml
@@ -24,8 +24,8 @@ spec:
         {{- if .Values.vaultAgent.enabled }}
         vault.hashicorp.com/agent-inject: "true"
         vault.hashicorp.com/role: "{{ .Values.vaultAgent.role }}"
-        vault.hashicorp.com/agent-inject-secret-alga-auth-key: "{{ .Values.vaultAgent.sharedSecretPath }}"
-        vault.hashicorp.com/agent-inject-template-alga-auth-key: |
+        vault.hashicorp.com/agent-inject-secret-alga_auth_key: "{{ .Values.vaultAgent.sharedSecretPath }}"
+        vault.hashicorp.com/agent-inject-template-alga_auth_key: |
           {{`{{- with secret "`}}{{ .Values.vaultAgent.sharedSecretPath }}{{`" -}}
           export ALGA_AUTH_KEY="{{ .Data.data.alga_auth_key }}"
           {{- end }}`}}
@@ -88,16 +88,6 @@ spec:
               echo "  NPM version: $(npm --version)"
               echo "  Working directory: $(pwd)"
               echo "  Source code commit: $(cd /app && git rev-parse --short HEAD 2>/dev/null || echo 'unknown')"
-              echo ""
-              {{- if .Values.vaultAgent.enabled }}
-              if [ -f /vault/secrets/alga-auth-key ]; then
-                source /vault/secrets/alga-auth-key
-                echo "Vault secrets loaded from /vault/secrets/"
-              fi
-              if [ -f /vault/secrets/server-secrets ]; then
-                source /vault/secrets/server-secrets
-              fi
-              {{- end }}
               echo "Dev pod is ready for debugging. Keeping container alive..."
               # Keep container running
               while true; do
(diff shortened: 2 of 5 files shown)


robertisaacs:
## 🤖 AI Code Review

Great work – this PR moves us a big step closer to using Vault as the single source of truth for secrets and cleans up some long-standing TODOs around `NEXTAUTH_SECRET` handling and the Helm charts.  The refactors look clean overall and I like the extra validation you added in `startupValidation.ts`.

Critical / blocking points
1. Inconsistent source of `NEXTAUTH_SECRET`
   • `startupValidation.ts` now **requires** the value to be present in the ENV, but  
   • `authorizationMiddleware` and `shared/utils/encryption.ts` read it **only** from the SecretProvider (no ENV fallback any more).  
   If the secret provider is down (or when running locally with `docker-compose up` where Vault is not started) we’ll fail auth even though the variable may be present.  
   → Either drop the ENV check or reinstate an ENV fallback so both places agree on the single canonical source.

2. `ALGA_AUTH_KEY` no longer ends up in the process environment
   • The previous Helm template wrote  
     `export ALGA_AUTH_KEY="..."` and the entrypoint sourced the file.  
   • That export + `source` lines have been removed, and the injected template now only writes the raw value.  
   Unless we set `env:` explicitly elsewhere (I couldn’t find it), the server will not have `process.env.ALGA_AUTH_KEY` and any internal API calls that rely on it will break.  
   → Please verify that `ALGA_AUTH_KEY` is still exposed as an environment variable inside the container, or change the code to read the mounted file path instead of `process.env`.

3. Secret key spelling change (`alga-auth-key` → `alga_auth_key`)
   The SecretProvider still uses the dashed key in a few places (`getAppSecret('ALGA_AUTH_KEY')`). Double-check the Vault secret JSON to make sure the field name change is intentional; otherwise lookups will start returning `undefined`.

Suggestions / nice-to-have
• The removal of the “defaultKey” fallback in `encryption.ts` is 👍 from a security standpoint, but consider a clearer error message hinting at “Have you run `pnpm dev:vault`?” for local devs.  
• You might want to add a health-check or startup probe that fails fast when SecretProvider can’t fetch `NEXTAUTH_SECRET`, so we don’t start pods that will 401 every request.  
• Since we’re touching the Helm templates, aligning the annotation blocks (`|` vs `|-`) would keep yamllint happy.

Overall this is a solid improvement; once the secret-source discrepancies are ironed out we’ll be good to merge.  Thanks for tightening up our secrets story!

---

RobertAtNineMinds:
> ## 🤖 AI Code Review
> Great work – this PR moves us a big step closer to using Vault as the single source of truth for secrets and cleans up some long-standing TODOs around `NEXTAUTH_SECRET` handling and the Helm charts. The refactors look clean overall and I like the extra validation you added in `startupValidation.ts`.
> 
> Critical / blocking points
> 
> 1. Inconsistent source of `NEXTAUTH_SECRET`
>    • `startupValidation.ts` now **requires** the value to be present in the ENV, but
>    • `authorizationMiddleware` and `shared/utils/encryption.ts` read it **only** from the SecretProvider (no ENV fallback any more).

codereview bot is missing: secret provider does read form ENV so no problem there.

>    If the secret provider is down (or when running locally with `docker-compose up` where Vault is not started) we’ll fail auth even though the variable may be present.
>    → Either drop the ENV check or reinstate an ENV fallback so both places agree on the single canonical source.
> 2. `ALGA_AUTH_KEY` no longer ends up in the process environment
>    • The previous Helm template wrote
>    `export ALGA_AUTH_KEY="..."` and the entrypoint sourced the file.
>    • That export + `source` lines have been removed, and the injected template now only writes the raw value.
>    Unless we set `env:` explicitly elsewhere (I couldn’t find it), the server will not have `process.env.ALGA_AUTH_KEY` and any internal API calls that rely on it will break.
>    → Please verify that `ALGA_AUTH_KEY` is still exposed as an environment variable inside the container, or change the code to read the mounted file path instead of `process.env`.

We don't source it because our secret provider has the ability to read the file.

> 3. Secret key spelling change (`alga-auth-key` → `alga_auth_key`)
>    The SecretProvider still uses the dashed key in a few places (`getAppSecret('ALGA_AUTH_KEY')`). Double-check the Vault secret JSON to make sure the field name change is intentional; otherwise lookups will start returning `undefined`.
> 

We're good here.

> Suggestions / nice-to-have • The removal of the “defaultKey” fallback in `encryption.ts` is 👍 from a security standpoint, but consider a clearer error message hinting at “Have you run `pnpm dev:vault`?” for local devs. • You might want to add a health-check or startup probe that fails fast when SecretProvider can’t fetch `NEXTAUTH_SECRET`, so we don’t start pods that will 401 every request. • Since we’re touching the Helm templates, aligning the annotation blocks (`|` vs `|-`) would keep yamllint happy.

Pretty sure we already do this. Once the AI review bot can do research, these types of comments will improve.

> 
> Overall this is a solid improvement; once the secret-source discrepancies are ironed out we’ll be good to merge. Thanks for tightening up our secrets story!
> 
> _Generated by AI Code Review System_ | [Review Configuration](https://github.com/Nine-Minds/nm-kube-config/blob/main/ai-code-review/ai-review-config.yaml)

PR #1302, opened 2025-11-19
delete button and active inactive functions

Janae-K:
Can delete client and contact in drawer, can switch client to inactive from popup if they have too much associated data, fixed lag time and bug for manual toggle.

### DIFF
--- a/server/src/context/TagContext.tsx
+++ b/server/src/context/TagContext.tsx
@@ -137,38 +137,39 @@ export const TagProvider = ({ children }: { children: ReactNode }) => {
   }, [refetchTagsStable]);
 
   // Use useRef to create a stable function that doesn't change on re-renders
-  // Define the function implementation
-  const getPermissionsImpl = async (entityType: TaggedEntityType): Promise<TagPermissions> => {
-    // Return cached permissions if available using ref to avoid stale closures
-    const currentPerms = permissionsRef.current[entityType];
-    if (currentPerms) {
-      return currentPerms;
-    }
-
-    // Check if there's already a pending request for this entityType
-    if (pendingPermissions.current[entityType]) {
-      return pendingPermissions.current[entityType]!;
-    }
-
-    // Create a new promise and cache it
-    const permissionPromise = checkTagPermissions(entityType)
-      .then(perms => {
-        setPermissions(current => ({ ...current, [entityType]: perms }));
-        return perms;
-      })
-      .finally(() => {
-        // Clear the pending promise
-        pendingPermissions.current[entityType] = null;
-      });
-
-    // Store the promise so other concurrent calls can use it
-    pendingPermissions.current[entityType] = permissionPromise;
-
-    return permissionPromise;
-  };
-
-  // Initialize ref with the implementation - ref ensures stable identity across renders
-  const getPermissionsRef = useRef(getPermissionsImpl);
+  const getPermissionsRef = useRef<(entityType: TaggedEntityType) => Promise<TagPermissions>>();
+  
+  if (!getPermissionsRef.current) {
+    getPermissionsRef.current = async (entityType: TaggedEntityType): Promise<TagPermissions> => {
+      // Return cached permissions if available using ref to avoid stale closures
+      const currentPerms = permissionsRef.current[entityType];
+      if (currentPerms) {
+        return currentPerms;
+      }
+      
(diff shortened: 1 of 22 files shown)


DELETION_RULES.md
```diff
@@ -0,0 +1,101 @@
...
+2. **Audit Compliance** - Maintains required business records
+3. **Legal Protection** - Preserves contracts, invoices, and communications
+4. **Business Continuity** - Protects operational history and reporting
+5. **User Experience** - Clear, predictable behavior that matches user expectations
+
+## 📋 **Best Practices**
+
+### Instead of Deleting:
+1. **Archive clients** with business history
+2. **Mark contacts as inactive** rather than deleting
```
NatalliaBukhtsik: what is the difference between 'archive' and 'mark as inactive'? there shoudn't be any, as it creates two parallel paths for same functionality

server/src/components/contacts/ClientContactsList.tsx
```diff
@@ -51,7 +51,7 @@ const ClientContactsList: React.FC<ClientContactsListProps> = ({ clientId, clien
       setLoading(true);
       setError(null);
       try {
-        const fetchedContacts = await getContactsByClient(clientId, 'active'); // Default to active
+        const fetchedContacts = await getContactsByClient(clientId, 'all'); // Show both active and inactive
```
NatalliaBukhtsik: Forces users to see inactive contacts with no way to hide them. Adding a filter/toggle would be a better option and more consistent with main contacts page

server/src/context/TagContext.tsx
```diff
@@ -202,14 +202,28 @@ export const useTags = () => {
...
     if (!context.tagsLoaded && !context.isLoading) {
       context.refetchTags();
     }
   }, [context]);
-  
+
   return context;
+};
+
+// Safe version that returns null if not within TagProvider (for optional usage)
```
NatalliaBukhtsik: This appears to be a workaround for TagManager crashing outside TagProvider context. The proper fix would be ensuring the component tree has TagProvider, not silently disabling tag functionality.

server/src/lib/actions/client-actions/clientActions.ts
```diff
@@ -73,16 +73,12 @@ export async function updateClient(clientId: string, updateData: Partial<Omit<IC
     throw new Error('No authenticated user found');
   }
```
NatalliaBukhtsik: Without this check, client updates bypass existing authorization safeguards. Restoring it preserves the current security model.

server/src/lib/actions/contact-actions/contactActions.tsx
```diff
@@ -121,9 +129,41 @@ export async function deleteContact(contactId: string) {
     const dependencies: string[] = [];
     const counts: Record<string, number> = {};
 
-    // Check for dependencies
+    // Check contracts first (outside main transaction to avoid abort issues)
```
NatalliaBukhtsik: Contract check is invalid - no contract table has contact_id. Contracts are linked to clients, not contacts. This query errors silently and never blocks anything.
And in general, a client having contracts shouldn't prevent deleting an unrelated contact.

server/src/lib/actions/contact-actions/contactActions.tsx
```diff
@@ -1596,17 +1791,17 @@ export async function updateContactPortalAdminStatus(
       throw new Error('Unauthorized');
     }
 
-    // Check permissions
```
NatalliaBukhtsik: same operations order change as above

server/src/lib/auth/rbac.ts
```diff
@@ -69,7 +69,37 @@ function canonicalizeResource(resource: string): string {
   return RESOURCE_CANONICAL_MAP[resource] ?? resource;
 }
 
-export async function hasPermission(user: IUser, resource: string, action: string, knexConnection?: Knex | Knex.Transaction): Promise<boolean> {
+// Overloaded signatures for hasPermission
```
NatalliaBukhtsik: unnecessary addition - using the existing signature keeps our auth code simple and consistent

NatalliaBukhtsik:
"This PR has merged main multiple times and contains unrelated features. Merging to release/0.15.0 will bring all of that in!!!

PR #1507, opened 2025-12-29
Onlyonedefaultclientlocationwhenaddingorediting

Janae-K:
Error messages consistent - all 4 error messages for location say "Location not found"
 Default logic consistent - all three functions (create, update, setDefault) now clear other defaults the same way.
 is_active check consistent - both updateClientLocation and setDefaultClientLocation require active location when setting default
 Adding and/or editing locations and switching default location toggle updates star default to only one main location.

### DIFF
--- a/server/src/lib/actions/client-actions/clientLocationActions.ts
+++ b/server/src/lib/actions/client-actions/clientLocationActions.ts
@@ -129,6 +129,7 @@ export async function updateClientLocation(
 
   const updatedLocation = await withTransaction(knex, async (trx: Knex.Transaction) => {
     // Get the existing location first to check current state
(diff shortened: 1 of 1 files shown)


NatalliaBukhtsik (changes requested):
When updating a location to set is_default = false, there's no validation that at least one default will remain (not sure if we always need at least one default location, but we reassign default flag to another location when deleting current default - let's add same logic here?)
The update query doesn't filter by is_active (we don't have options to set a location as inactive at all, but if we're filtering only active everywhere - should be consistent)

NatalliaBukhtsik (approved):
merging it in!

PR #1509, opened 2025-12-29
fix(time-management): fix period column sorting in timesheet approvals

Janae-K:
## Summary
- Fix period column sorting in timesheet approvals dashboard
- Change dataIndex from `time_period` to `time_period.start_date` to enable proper date sorting
- Add `sortable: false` to Select and Actions columns for consistency

## Root Cause
The period column was not sorting because the entire `time_period` object was being passed to the sort function, which converted to `"[object Object]"` for all rows, making all rows appear equal.

## Test plan
- [ ] Navigate to Time Sheet Approvals page
- [ ] Click on the Period column header to sort
- [ ] Verify sorting works correctly (ascending/descending by start date)
- [ ] Verify Select and Actions columns no longer show sort arrows

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "@blocknote/core": "^0.22.0",
     "@blocknote/mantine": "^0.22.0",
     "@blocknote/react": "^0.22.0",
-    "@hello-pangea/dnd": "^16.6.0",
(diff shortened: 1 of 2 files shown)


NatalliaBukhtsik (approved):
comments could be less explicit and wordy, but looks good to me

PR #1524, opened 2026-01-01
When in client view, contact view, etc., deleting documents should make more clear that it deletes the document from the ENTIRE PSA and it will be removed from all associated items that have that document attached

robertisaacs:
## Summary
- Changed "Delete" button to "Delete Permanently" to emphasize the action's permanence
- Updated delete confirmation message to explain that the file will be removed entirely from the system and won't be available anywhere
- Changed "Remove" button to "Detach" to clarify that it only removes the link
- Updated detach confirmation message to explain that the document remains in the library and can be attached to other items

## Test plan
- [ ] Navigate to a client view with attached documents
- [ ] Click the menu on a document and verify "Delete Permanently" and "Detach" options are shown
- [ ] Click "Delete Permanently" and verify the confirmation dialog explains system-wide removal
- [ ] Click "Detach" and verify the confirmation dialog explains only the link is removed
- [ ] Test the same for video documents


NatalliaBukhtsik (changes requested):
If changing language keys please update corresponding values in language packs

NatalliaBukhtsik (approved):

PR #1606, opened 2026-01-13
save button and popup unsaved changes following contracts and projects

Janae-K:
Purple checkmark and x for title changes, block save changes button to mimic billing contracts, unsaved changes continue editing and discard changes popup to mimic projects behavior. 

-Dropdowns and just have them rely on bottom save button. 
-No other save changes button for blocks besides the top section. Save changes button on bottom of ticket block to be identical in copy same behavior is used in contracts. Title have it copy what is used in title/name for contracts with purple checkmark and x like contracts but if you press check button, it will save the title regardless of save changes. Also, if you edit the title and don’t click anything but you click save changes at bottom, it will still save changes you made to title and all else. Aka 2 saves. 
-If you change the board to itil board, all fields need to pop up/all options for user needs to show, even if you dont click save when changing board. Same when switching back from itil to a diff board. 
-Make sure tickets don't change board when a user changes the dropdown unless they click save changes. 
-Add confirmation dialogue if user clicks out of title or dropdowns or etc without clicking save buttons if changes made and also not clicking save changes button below. Both dialogue and pop-up like what is already used on our platform like projects, all projects has. Use what we have when tracking unsaved changes like billing contracts. and just keep continue editing and etc like contracts.


NatalliaBukhtsik (approved):
ready to be merged

PR #1736, opened 2026-02-02
Fix notifications mark all read not working

arynshimmy:
## Summary

Fixed a bug where clicking "Mark all read" in the notifications dropdown would not actually mark notifications as read.

**Root cause:** The `useInternalNotifications` hook in `packages/notifications/src/hooks/useInternalNotifications.ts` was calling the server actions with object arguments:

```typescript
await markAllAsReadAction({ tenant, user_id: userId });
await markAsReadAction({ tenant, user_id: userId, notification_id: notificationId });
```

However, the server actions in `internalNotificationActions.ts` expect positional arguments:

```typescript
export async function markAllAsReadAction(tenant: string, userId: string)
export async function markAsReadAction(tenant: string, userId: string, notificationId: string)
```

This mismatch caused the `tenant` parameter to receive the entire object, and `userId` to be `undefined`. The database query then failed with "Undefined binding(s) detected when compiling UPDATE. Undefined column(s): [user_id]".

**Fix:** Changed the hook to pass positional arguments:

```typescript
await markAllAsReadAction(tenant, userId);
await markAsReadAction(tenant, userId, notificationId);
```

Note: The duplicate hook in `server/src/hooks/useInternalNotifications.ts` already had the correct implementation.

### DIFF
--- a/packages/core/src/lib/validation.ts
+++ b/packages/core/src/lib/validation.ts
@@ -4,6 +4,8 @@ import { Temporal } from '@js-temporal/polyfill';
 
 // Basic validation utilities
 
+const basicEmailSchema = z.string().email();
+
 /**
  * Simple email format validation - null-safe wrapper
  * Use this for quick email checks before sending notifications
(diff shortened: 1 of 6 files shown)


NatalliaBukhtsik (approved):
Remove the // @ts-nocheck directive from the file now that the function calls are fixed. (was excluded earlier to exclude from typecheck to bypass the check instead of fixing it)

RobertAtNineMinds (approved):

PR #1813, opened 2026-02-09
Restore auto-tracked interval display and fix duration bug

arynshimmy:
## Summary

- **Restore the IntervalManagement component** in the ticket properties sidebar, which was accidentally removed during the modularity architecture refactor. This re-adds the "Tracked Intervals" section below the ticket timer, showing auto-tracked time intervals with their durations, start/end times, active/auto-closed badges, and actions to select, merge, delete, or convert intervals into time entries.

- **Fix a unit mismatch bug in IntervalTrackingService** where `endInterval()` and `cleanupOrphanOpenIntervals()` stored interval duration in **milliseconds** while all display code (`formatDuration`, `IntervalItem`, `calculateTotalDuration`) treats the `duration` field as **seconds**. This caused intervals to display ~1000x their actual length (e.g. a 1-minute interval appeared as ~18 hours). The `autoCloseOpenIntervals()` method was already correctly dividing by 1000, confirming this was an oversight in the other two methods.

- **Use the render prop composition pattern** to inject IntervalManagement into TicketProperties, following the existing architecture used for `renderContactDetails`, `renderClientDetails`, etc. The `@alga-psa/tickets` package never imports from `@alga-psa/scheduling` directly — the wiring happens in `MspTicketDetailsContainerClient` (the composition layer).

All changes are in a single commit.

### `packages/ui/src/services/IntervalTrackingService.ts`
- `endInterval()`: Change `endTime.getTime() - startTime.getTime()` to `Math.floor((endTime.getTime() - startTime.getTime()) / 1000)` so duration is stored in seconds
- `cleanupOrphanOpenIntervals()`: Same fix — divide by 1000 to store seconds instead of milliseconds

### `packages/tickets/src/components/ticket/TicketProperties.tsx`
- Add optional `renderIntervalManagement` render prop to interface
- Render it inside the Time Entry card below the "Add Time Entry" button, gated on `ticket.ticket_id && userId && renderIntervalManagement`

### `packages/tickets/src/components/ticket/TicketDetails.tsx`
- Add `renderIntervalManagement` to `TicketDetailsProps` interface
- Destructure and pass through to `TicketProperties`

### `packages/tickets/src/components/ticket/TicketDetailsContainer.tsx`
- Accept and forward `renderIntervalManagement` prop

### `packages/msp-composition/src/tickets/MspTicketDetailsContainerClient.tsx`
- Import `IntervalManagement` from `@alga-psa/scheduling`
- Create `renderIntervalManagement` callback and pass to `TicketDetailsContainer`
- Add to the `Omit` type so consumers don't need to provide it

## Note
Existing intervals stored in IndexedDB prior to this fix will still have inflated millisecond-based duration values. Users can delete those stale intervals through the UI. All new intervals going forward will have correct durations.

## Test plan
- [ ] Open a ticket and verify the "Tracked Intervals" section appears below the "Add Time Entry" button
- [ ] Stay on the ticket for a short period, navigate away, then return — verify the interval duration displays correctly (seconds, not inflated to hours)
- [ ] Test selecting intervals and using merge/delete/convert-to-time-entry actions
- [ ] Verify the ticket timer (start/pause/reset) still functions correctly alongside the interval display

### DIFF
--- a/packages/msp-composition/src/tickets/MspTicketDetailsContainerClient.tsx
+++ b/packages/msp-composition/src/tickets/MspTicketDetailsContainerClient.tsx
@@ -7,12 +7,13 @@ import ClientDetails from '@alga-psa/clients/components/clients/ClientDetails';
 import type { IClient, IContact } from '@alga-psa/types';
 import CreateTaskFromTicketDialog from '@alga-psa/projects/components/CreateTaskFromTicketDialog';
 import LinkTicketToTaskDialog from '@alga-psa/projects/components/LinkTicketToTaskDialog';
+import { IntervalManagement } from '@alga-psa/scheduling/components/time-management/interval-tracking/IntervalManagement';
 import { TicketIntegrationProvider } from '@alga-psa/projects/context/TicketIntegrationContext';
 import { useTicketIntegrationValue } from '../projects/useTicketIntegrationValue';
 
 type MspTicketDetailsContainerClientProps = Omit<
   React.ComponentProps<typeof TicketDetailsContainer>,
-  'renderContactDetails' | 'renderClientDetails'
+  'renderContactDetails' | 'renderClientDetails' | 'renderIntervalManagement'
 >;
 
 export default function MspTicketDetailsContainerClient(props: MspTicketDetailsContainerClientProps) {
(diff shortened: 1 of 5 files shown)


NatalliaBukhtsik (changes requested):
importing IntervalManagement directly from scheduling into tickets — bypasses this composition layer. The proper pattern would be:

1. tickets/TicketProperties accepts an optional renderIntervalManagement render prop or reads from a context
2. msp-composition/MspTicketDetailsContainerClient provides the IntervalManagement component from scheduling via that prop/context
3. tickets package never knows about scheduling and doesn't create circular dependency

PR #2144, opened 2026-03-24
Add project task CSV export

arynshimmy:
## Summary
- Adds an "Export Tasks" button to the project detail header (next to Save as Template, Materials, Edit)
- Opens an export dialog where users can select which phases and which fields to include in the CSV
- Server action fetches tasks for selected phases, resolves names (users, teams, statuses, priorities, task types), and generates CSV with proper escaping and injection prevention

## Test plan
- [ ] Navigate to a project detail page and verify the "Export Tasks" button appears in the header
- [ ] Click the button and verify the dialog opens with all phases and fields selected by default
- [ ] Deselect some phases and fields, then export — verify the CSV only contains the selected phases/fields
- [ ] Verify select/deselect all toggles work for both phases and fields
- [ ] Verify the exported CSV opens correctly in a spreadsheet application
- [ ] Verify the export button is disabled when no phases or no fields are selected

### DIFF
--- a/packages/projects/src/actions/projectTaskExportActions.ts
+++ b/packages/projects/src/actions/projectTaskExportActions.ts
@@ -1,0 +1,398 @@
+'use server'
+
+import type { ITag } from '@alga-psa/types';
+import { createTenantKnex, withTransaction } from '@alga-psa/db';
+import { withAuth } from '@alga-psa/auth';
+import { hasPermission } from '@alga-psa/auth/rbac';
+import { findTagsByEntityIds } from '@alga-psa/tags/actions';
+import { Knex } from 'knex';
+
+const MAX_EXPORT_ROWS = 10000;
+
+const CSV_FIELDS = [
+  'task_name',
+  'description',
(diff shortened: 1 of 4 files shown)


packages/projects/src/actions/projectTaskExportActions.ts
```diff
@@ -0,0 +1,398 @@
...
+  projectId: string,
+  selectedPhaseIds: string[],
+  selectedFields?: string[],
+): Promise<{ csv: string; count: number }> => {
+  const { knex: db } = await createTenantKnex();
+
+  return await withTransaction(db, async (trx: Knex.Transaction) => {
+    const hasRead = await hasPermission(_user, 'project', 'read', trx);
+    if (!hasRead) {
+      throw new Error('Permission denied: Cannot read project');
```
NatalliaBukhtsik: could use throwPermissionError from errorHandling utility

NatalliaBukhtsik (approved):

PR #2148, opened 2026-03-24
Fix workflow scheduler UI: select overflow and next fire time display

arynshimmy:
## Summary

- **CustomSelect overflow fix**: Workflow options in the scheduler dialog were overflowing and misaligning the select trigger when eligibility messages (e.g. "No published version") were present. Added a `dropdownHint` prop to `CustomSelect` so hints render only in the dropdown list, not the trigger. Flattened workflow option labels to single-line with truncation.

- **Next fire time for recurring schedules**: The "Next Fire / Run At" column only showed the raw cron expression (e.g. `43 13 * * *`) with a dash for the timestamp. Now the server computes the actual next fire time using `cron-parser` and persists it to the existing `next_fire_at` DB column on every create, update, restore, enable/disable, and after each successful recurring run. Supports the full 5-field cron spec including steps (`*/15`), ranges (`1-5`), and month filters (`*/2`).

- **Timestamp format alignment**: Changed the schedule list's timestamp display from `toLocaleString()` (which included seconds) to `Intl.DateTimeFormat` with `dateStyle: 'medium', timeStyle: 'short'`, matching the rest of the app.

## Files changed

| File | Change |
|------|--------|
| `packages/ui/.../CustomSelect.tsx` | Add `dropdownHint` prop for dropdown-only hint text |
| `ee/.../WorkflowScheduleDialog.tsx` | Use `dropdownHint` for eligibility messages, flatten option layout |
| `ee/.../computeNextFireAt.ts` | New utility wrapping `cron-parser` for timezone-aware next fire computation |
| `ee/.../workflowScheduleLifecycle.ts` | Populate `next_fire_at` on all schedule persist paths |
| `server/.../workflowScheduledRunHandlers.ts` | Update `next_fire_at` after each recurring run fires |
| `ee/.../Schedules.tsx` | Use `dateStyle: 'medium', timeStyle: 'short'` format |
| `ee/.../workflows/package.json` | Add explicit `cron-parser` dependency |

## Test plan

- [ ] Create a recurring schedule via the scheduler dialog — verify "Next Fire / Run At" shows a human-readable date (e.g. "Mar 25, 2026, 1:43 PM") instead of "—"
- [ ] Verify the cron expression + timezone still displays below the timestamp
- [ ] Edit a recurring schedule's cron/timezone — verify the next fire time updates
- [ ] Pause and resume a schedule — verify next fire time clears on pause and recomputes on resume
- [ ] Create a one-time schedule — verify "Next Fire / Run At" shows the run_at time
- [ ] Verify workflow select in the scheduler dialog no longer overflows with long workflow names or eligibility messages
- [ ] Run existing handler tests: `npx vitest run src/test/unit/workflowScheduledRunHandlers.unit.test.ts` (3/3 passing)

### DIFF
--- a/ee/packages/workflows/package.json
+++ b/ee/packages/workflows/package.json
@@ -100,6 +100,7 @@
     "@alga-psa/validation": "*",
     "ajv": "^8.12.0",
     "ajv-formats": "^2.1.1",
+    "cron-parser": "^4.9.0",
     "zod": "^3.22.4",
     "zod-to-json-schema": "^3.24.6"
   },
--- a/ee/packages/workflows/src/components/automation-hub/Schedules.tsx
+++ b/ee/packages/workflows/src/components/automation-hub/Schedules.tsx
@@ -72,7 +72,7 @@ const formatTimestamp = (value?: string | null): string => {
   if (!value) return '—';
   const date = new Date(value);
   if (Number.isNaN(date.getTime())) return '—';
-  return date.toLocaleString();
+  return new Intl.DateTimeFormat(undefined, { dateStyle: 'medium', timeStyle: 'short' }).format(date);
 };
 
 const formatRelativeTimestamp = (value?: string | null): string => {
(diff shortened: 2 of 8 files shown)


RobertAtNineMinds (approved):
looks good

PR #2214, opened 2026-04-01
feat: show time entry history on ticket detail view

iamJeb:
Closes #2152

Time entries logged from a ticket weren't visible on the ticket itself — you had to go to Time Management > Time Sheet to see them. The ticket page never queried `time_entries` and had nowhere to display them.

This adds a "Time Entries" tab to the ticket conversation area that shows all entries for that ticket: who logged it, when, duration, service, billable status, and approval status. New entries show up immediately after saving. Tab is MSP portal only, not exposed to client portal.

Doesn't cover the optional enhancements from the issue (time summary totals, inline editing, auto-generated comments) — those can be follow-ups.

### DIFF
--- a/packages/scheduling/src/actions/timeEntryActions.ts
+++ b/packages/scheduling/src/actions/timeEntryActions.ts
@@ -12,6 +12,7 @@ import {
 } from './timeSheetOperations';
 import {
   fetchTimeEntriesForTimeSheet,
+  fetchTimeEntriesForWorkItem,
   saveTimeEntry,
   updateTimeEntryApprovalStatus,
   deleteTimeEntry,
@@ -39,6 +40,7 @@ export {
   fetchTimePeriods,
   fetchOrCreateTimeSheet,
   fetchTimeEntriesForTimeSheet,
+  fetchTimeEntriesForWorkItem,
   saveTimeEntry,
   updateTimeEntryApprovalStatus,
   deleteTimeEntry,
--- a/packages/scheduling/src/actions/timeEntryCrudActions.ts
+++ b/packages/scheduling/src/actions/timeEntryCrudActions.ts
@@ -251,6 +251,8 @@ export const fetchTimeEntriesForTimeSheet = withAuth(async (
   return attachTimeEntryChangeRequests(entriesWithWorkItems, changeRequestsByEntryId);
 });
 
+const ALLOWED_WORK_ITEM_TYPES = new Set(['ticket', 'project_task', 'non_billable_category', 'ad_hoc', 'interaction']);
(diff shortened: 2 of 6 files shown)


NatalliaBukhtsik (changes requested):
Thanks for your submission! However, there are several issues with this PR:

1. Wrong location — time entries don't belong in conversationtabs

The conversation area is for comments/communication history. The ticket already has a "Time Entry" card in the right sidebar (TicketProperties.tsx:381-448) with timer controls and "Add Time Entry" button.If we want to show time entry history, it should be added to that card — not as a conversationtab.

2. Use existing UIcomponents

The PR uses raw <span> elements with hardcoded Tailwind color classes for badges. We have a Badge component at packages/ui/src/components/Badge.tsx with propervariants:
- Approval status: use Badge with variant="success" (APPROVED), variant="warning" (SUBMITTED), variant="error" (CHANGES_REQUESTED). This is the exact pattern used in TimeSheetApprovaltsx.
- Billable/Non-billable: use Badge with variant="info" /variant="default".
- Card containers: use Card from packages/ui/src/components/Card.tsx instead of raw divs with borderclasses.

3. Invalid Date bug
 
The date formatting:
new Date(entry.work_date + 'T00:00:00').toLocaleDateStrin()
work_date is a YYYY-MM-DD string (generated via Temporal.PlainDate.toString()). Appending 'T00:00:00' without a timezone creates ambiguous parsing across browsers. The codebase uses date-fns everywhere else— the correct patternis:
import { parseISO, format } from'date-fns';
format(parseISO(entry.work_date), 'MM/ddyyyy')
There's also formatMinutesAsHoursAndMinutes() in packages/core/src/lib/dateTimeUtils.ts for the durationdisplay.

4. Access control — technicians can see other users'entries

This is the most critical issue. The query in optimizedTicketActions.ts fetches all time entries for the ticket and only gates on hasPermission(user, 'timeentry', 'read'). Every other time entry endpointuses
assertCanActOnBehalf() from packages/scheduling/src/actions/timeEntryDelegationAuth.ts to verify the requesting user has delegation authority over the entry owner. Withoutthis:

- A technician sees all other users' time entries, notes, and billable amounts on theticket                                                                                       
- This bypasses the established access controlmodel

At minimum, entries should be filtered to the current user's own entries unless they have timesheet:read_all permission or manage the entryowner.

5. Deadcode fetchTimeEntriesForWorkItem in timeEntryCrudActions.ts is added and exported but never called anywhere. Should be removed from this PR.

NatalliaBukhtsik:
already implemented in the time entry card

PR #2266, opened 2026-04-08
Add rich text editor for project task descriptions

arynshimmy:
## Summary
- Replace plain TextArea with BlockNote rich text editor for task descriptions (create and edit modes), with @mention support
- Add `taskRichText.ts` utility module for parsing, serializing, and extracting plain text from rich text descriptions
- Update all downstream consumers (board cards, list view, search, CSV export, template views/forms/copy) to extract plain text so stored BlockNote JSON never surfaces as raw text
- Preserve rich text formatting when creating a task from a ticket (both now use BlockNote)
- Extend rich text editing to project template tasks (TemplateTaskForm) so save-as-template and create-project-from-template no longer strip formatting
- Handle backward compatibility: legacy plain-text descriptions parse cleanly into the editor and don't trigger false "unsaved changes" detection

## Code-review fixes
- **No data loss in template flow**: `createTemplateFromProject` now writes `task.description` verbatim instead of stripping through `extractTaskDescriptionText`; `TemplateTaskForm` uses the same BlockNote `TextEditor` as `TaskForm`, so rich text round-trips through templates
- **`isTaskRichTextEmpty` recognizes non-text blocks**: image/table/video/audio/file/codeBlock/embed blocks are now treated as inherently non-empty, preventing image-only descriptions from being saved as `null`
- **Dirty-check false positives**: `TaskForm` and `TemplateTaskForm` now capture the BlockNote-normalized baseline via `editorRef` after mount and compare the live `editor.document` in `hasChanges()`, so BlockNote's added block IDs / default props no longer flag spurious unsaved changes
- **Format detection**: replaced `trimmed.startsWith('[')` with a `/^\s*\[\s*\{/` signature in `parseTaskRichTextContent` / `extractTaskDescriptionText` so plain text like `[URGENT] Fix bug` skips the wasted JSON.parse and the latent bug where valid-JSON-array-looking text would break parsing
- **Repeated JSON.parse per render**: `TaskListView` memoizes a `task_id → plain text` map keyed on `tasks`, so search auto-expand, the `filteredTasks` filter, and per-row rendering all share one parse per task instead of re-parsing 4× per task per render

## Test plan
- [x] Create a new project task — verify the rich text editor appears with formatting toolbar
- [x] Type formatted text (bold, italic, lists) and save — reopen to confirm formatting persists
- [x] Edit an existing task with a plain-text description — verify it loads cleanly and doesn't show unsaved changes on cancel
- [x] Use @mentions in the description — verify individual users appear (not just `@everyone`)
- [x] Check board view and list view show plain-text previews of rich text descriptions
- [x] Search for text within a rich-text description and verify it matches
- [x] Export tasks to CSV and verify descriptions export as plain text
- [x] Create a task from a ticket with formatted description — verify formatting carries over
- [ ] Save a project with formatted task descriptions as a template — verify rich text survives the round-trip
- [ ] Create a project from a template with rich text task descriptions — verify formatting is preserved
- [ ] Edit a template task in TemplateTaskForm — verify the BlockNote editor loads and no false "unsaved changes" prompt on cancel
- [ ] Add an image-only description to a task and save — verify it is not stored as `null`
- [ ] Verify a task description starting with `[` (e.g. `[URGENT] Fix bug`) loads without parse errors


NatalliaBukhtsik (changes requested):
- No rich text formatting in project template task descriptions -> data loss on save as template and create project from this template

- isTaskRichTextEmpty ignores non-text blocks — If a description contains only an image, table, or embed (blocks without a content array), the function returns true and the description gets saved as null. That's actual data loss outside of templates.

- Dirty-check false positives — The "unsaved changes" detection in TaskForm compares serializeTaskRichTextContent(descriptionContent) (which went through BlockNote normalization) against serializeTaskRichTextContent(parseTaskRichTextContent(task.description)) (which didn't). If BlockNote adds default props or reorders keys, opening a task and changing nothing will still trigger "you have unsaved changes." Annoying UX, not data loss.

- startsWith('[') format detection — A plain text description like [URGENT] Fix the login bug would be fed to JSON.parse, which would fail and fall back gracefully. So it's not a bug per se — but it's a wasted parse attempt on every render for those descriptions, and it would become a real bug if someone wrote a description that happened to be valid JSON array syntax.

- Repeated JSON.parse per render — TaskListView calls extractTaskDescriptionText up to 4 times per task in the same render pass. Each call re-parses the JSON. Not a correctness issue, but measurable on large task lists (200+ tasks).

packages/projects/src/components/project-templates/wizard-steps/TemplateTasksStep.tsx
```diff
@@ -1,6 +1,7 @@
 'use client';
 
 import React, { useState } from 'react';
+import { extractTaskDescriptionText } from '../../../lib/taskRichText';
 import { Label } from '@alga-psa/ui/components/Label';
 import { Input } from '@alga-psa/ui/components/Input';
 import { TextArea } from '@alga-psa/ui/components/TextArea';
```
NatalliaBukhtsik, line 15: no rich text editing in creating new templates from wizard

NatalliaBukhtsik (changes requested):
TemplateTaskForm uses undefined for empty descriptions — When rich text is empty, descriptionSerialized is undefined. If the onSave handler does a partial merge/spread, undefined won't overwrite the existing value — meaning clearing a template task description might not actually clear it in the DB. Compare with TaskForm which uses null explicitly. (This matches the pre-existing behavior from description.trim() || undefined, so it's not a regression — but it might be a latent bug worth verifying.)

NatalliaBukhtsik (changes requested):
now it will be trying to extractTaskDescriptionText from markdown expecting rich text there
and migration, of course (but that's small)

NatalliaBukhtsik (approved):

PR #2411, opened 2026-04-27
Improve project task import: fallback pickers, phase-scoped status resolution, better preview

arynshimmy:
## Summary

Closes several gaps in the project CSV phase/task import flow that surfaced while reviewing the import wizard.

### Fallback pickers (preview step)
- **Default phase for tasks without phase_name** — picker lists every existing phase on the project plus a "Create new 'Unsorted Tasks' phase" option. Only appears when at least one valid row has no phase_name. Changing the selection re-groups the preview live.
- **Default status for tasks without status** — picker filters to statuses scoped to the chosen default phase (phase-specific mappings if the phase has any, otherwise project-wide). Replaces the previous silent "first status" default.
- `groupRowsIntoPhases` accepts a new `defaultPhaseName` parameter; `importPhasesAndTasks` accepts a new `defaultStatusMappingId` parameter.

### Phase-scoped status resolution
- `IUnmatchedStatusInfo` and `IStatusResolution` now carry `phaseName`. The validation response returns `Array<{phaseName, statusName}>` instead of a flat string array, so the same status name appearing under two phases produces two resolution entries.
- Resolution step renders one card per (phase, status) pair, labels the affected phase with a branded `Badge`, and filters the "Map to existing" dropdown to that phase's available statuses.
- `importPhasesAndTasks` applies resolutions through a `phase → status → mappingId` map, so a task in Phase A gets Phase A's resolution and a task in Phase B gets Phase B's even when the CSV status string is identical.

### Row Results preview improvements
- New **Status** column showing the resolved status per row using phase-aware lookup. Three states: matched (plain), unmatched (warning `Badge` with tooltip directing users to the mapping steps), fallback (default `Badge` indicating the user-selected default will be applied).
- Existing valid/invalid icon column renamed to **Valid** so the two columns aren't both labeled "Status".
- New **row filter** dropdown: All / Will be imported / Will be skipped, with live counts.
- New warning alert when "Skip invalid rows" is enabled showing how many rows will be dropped.
- Indicators use the existing `Badge` component variants (`warning`, `default-muted`) wired to brand color tokens — no hardcoded Tailwind palette values.

### Error reporting
- `ITaskImportRow` and `IGroupedTaskData` carry an optional `csvRowNumber` so the origin row survives the validate → group → import pipeline.
- Task and phase creation errors now include the originating CSV row number(s).

### Reference data
- `getImportReferenceData` returns the project's existing phases (used to populate the fallback phase picker without an extra fetch). `IImportReferenceData` gains a `phases` field.

## Test plan

- [ ] Import a CSV where some rows omit `phase_name`. Confirm the preview shows the **Default phase** picker, listing existing phases plus "Create new 'Unsorted Tasks'", and that switching it re-groups the preview live.
- [ ] Import a CSV where some rows omit `status`. Confirm the **Default status** picker is scoped to the chosen default phase's statuses and that the chosen status is applied to those tasks on import.
- [ ] Import a CSV where the same status string (e.g. "Test") appears in two different phases that have different status mappings. Confirm two cards appear in the resolution step, each filtered to its phase's mappings, and that the resolutions are applied per-phase on import.
- [ ] Confirm the **Status** column in Row Results shows matched/unmatched/default states correctly with branded badges.
- [ ] Toggle "Skip invalid rows" with invalid rows present — confirm the warning alert reports the count and the "Will be skipped" filter shows them.
- [ ] Trigger a task creation error (e.g., constraint violation) and confirm the resulting error message includes the CSV row number.

### DIFF
--- a/packages/projects/src/components/PhaseTaskImportDialog.tsx
+++ b/packages/projects/src/components/PhaseTaskImportDialog.tsx
@@ -63,6 +63,57 @@ const MAX_IMPORT_ROWS = 5000;
 // Threshold for showing confirmation before import
 const LARGE_IMPORT_THRESHOLD = 100;
 
+/**
+ * Recompute the set of unmatched (phase, status) pairs given the current fallback phase.
+ *
+ * The fallback phase determines the effective phase for rows whose phase_name is blank,
+ * which in turn determines which phase-scoped status set their statuses are looked up in.
+ * Mirrors the server-side logic in validatePhaseTaskImportDataWithReferenceData so that
+ * the preview stays consistent if the user changes the fallback phase.
+ */
+function computeUnmatchedStatusState(
+  validationResults: ITaskImportValidationResult[],
+  defaultPhaseName: string,
+  statusLookup: Record<string, string>,
+  statusLookupByPhase: Record<string, Record<string, string>>
+): {
+  unmatchedStatuses: Array<{ phaseName: string; statusName: string }>;
+  unmatchedStatusInfo: IUnmatchedStatusInfo[];
+} {
+  const fallbackPhase = defaultPhaseName?.trim() || DEFAULT_PHASE_NAME;
+  const pairs = new Map<string, { phaseName: string; statusName: string }>();
+  const infoMap = new Map<string, IUnmatchedStatusInfo>();
+
+  for (const result of validationResults) {
+    const statusName = result.data.status?.trim();
+    if (!statusName) continue;
+    const phaseName = result.data.phase_name?.trim() || fallbackPhase;
+    const lookup = statusLookupByPhase[phaseName.toLowerCase()] || statusLookup;
+    if (lookup[statusName.toLowerCase()]) continue;
+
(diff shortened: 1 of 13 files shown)


packages/projects/src/components/PhaseTaskImportDialog.tsx
```diff
@@ -812,6 +1028,62 @@ const PhaseTaskImportDialog: React.FC<PhaseTaskImportDialogProps> = ({
                   />
                 </div>
               </div>
+
+              {/* Fallback phase picker — only shown when at least one valid row has no phase */}
+              {unphasedRowCount > 0 && (
+                <div className="flex items-center justify-between py-3 border-b">
+                  <div>
+                    <div className="text-gray-900 font-medium">
+                      {importT('fallbackPhaseLabel', 'Default phase for tasks without phase_name')}
```
NatalliaBukhtsik, line 1037: these need to go into each locale json files to be translated correctly

NatalliaBukhtsik (changes requested):
please add translations to ../features/projects.json

NatalliaBukhtsik (approved):

PR #2434, opened 2026-04-29
fix(contract-wizard): preview frequency, summary placement, and field layouts

arynshimmy:
## Summary

A bundle of contract-creation wizard polish fixes — one functional bug in the recurring preview engine, plus several layout/UX fixes across the wizard steps.

### Recurring preview honors all billing frequencies (functional bug)

The shared `recurringAuthoringPreview` helper had `resolvePreviewBillingFrequency()` whitelisting only `quarterly` and `annually`. Every other value — `weekly`, `bi-weekly`, `semi-annually`, even `monthly` — fell through the `default` branch and was silently coerced to `monthly`. This meant the **Illustrative future materialized periods** preview always showed monthly cycles regardless of what alternate frequency was selected. The bug affected **all four** call sites: contract-wizard Fixed Fee step, contract-wizard Review step, template-wizard Fixed Fee step, template-wizard Review step.

**Fix:** split frequency resolution by cadence owner so each call passes through every value the underlying materializer actually supports:
- `resolveClientPreviewBillingFrequency` → full `BillingCycleType`, matching `materializeClientCadenceServicePeriods`.
- `resolveContractPreviewBillingFrequency` → `monthly | quarterly | semi-annually | annually`, falling back to `monthly` for `weekly` / `bi-weekly` (contract-cadence engine constraint).
- `buildMaterializedPreviewPeriods` now takes the raw frequency string and resolves per-cadence at call time.

### PO Amount input rendered nested boxes
The wrapped Input always emitted its own `border` and `focus:ring-2`, which Tailwind source order kept in effect even when overrides tried to suppress them — so the inner input drew a second border + focus ring inside the outer wrapper. Replaced the nested Input with a plain native `<input>` styled transparent inside the existing bordered wrapper. Also gave the wrapper the background color and `flex-1 min-w-0` to the input so it grows next to the currency symbol.

### Fixed Fee step
- Recurring Preview Before Save panel now renders **below** the Alternate Billing Frequency selector instead of above it (it was previously appearing before the control that drove its values).
- Summary now includes an **Alternate Billing Frequency** line when an override differs from the contract's frequency.
- Added a tooltip (HelpCircle) next to **Illustrative future materialized periods** explaining what those rows represent — preview of the next few service periods and invoice windows that would be generated under current settings, for sanity-checking the cadence; actual invoices are produced later by the billing run.

### Hourly Services step
- Hourly Services Summary moved to **below** the Alternate Billing Frequency selector.
- Summary now also lists each bucket-enabled service with **Included Hours** (`total_minutes / 60`), **Overage Rate** (formatted from `overage_rate` cents), and **Rollover** status.
- When an alternate billing frequency is set, the summary shows it on its own line.

### Layout alignment
- **Products step (Quantity / Override unit price):** row used `items-end`, which aligned column bottoms. Only the Override column has trailing helper text (`Default catalog price: $X`), so its input was pushed up. Switched to `items-start` so inputs align at the top.
- **Usage-Based step (Rate per Unit / Unit of Measure):** Rate per Unit label has a Coins icon in a flex container while Unit of Measure was plain text, creating a small vertical offset between the two columns. Both labels now use `flex items-center gap-2 h-5` so they render at a consistent fixed height with or without an icon.

## Test plan

- [ ] PO Amount field shows a single bordered input with the contract's currency symbol as the prefix (e.g., $, €, £); focusing renders one ring, not nested ones.
- [ ] In Fixed Fee step, set Alternate Billing Frequency to **Weekly**; the Illustrative future materialized periods list shows ~7-day service periods (e.g. `2025-04-01 to 2025-04-08`), not monthly windows.
- [ ] Try **Bi-weekly** and **Semi-annually** as alternate frequencies and verify the periods reflect those cadences too.
- [ ] When an alternate billing frequency is selected, both the Fixed Fee and Hourly summaries show the **Alternate Billing Frequency** line.
- [ ] Recurring Preview panel and Hourly Services Summary now render **after** the Alternate Billing Frequency selector, not before it.
- [ ] Hover the help icon next to **Illustrative future materialized periods** — tooltip explains what the rows represent.
- [ ] Hourly Services Summary shows **Included Hours** and **Overage Rate** for any service with bucket allocation enabled.
- [ ] In Products step, Quantity and Override unit price inputs sit on the same vertical line.
- [ ] In Usage-Based step, Rate per Unit and Unit of Measure inputs sit on the same vertical line.
- [ ] Confirm template-wizard Fixed Fee and Review steps also render correct preview periods for non-monthly/non-quarterly/non-annual frequencies.

### DIFF
--- a/packages/billing/src/components/billing-dashboard/contracts/recurringAuthoringPreview.ts
+++ b/packages/billing/src/components/billing-dashboard/contracts/recurringAuthoringPreview.ts
@@ -32,15 +32,41 @@ export type RecurringAuthoringPreview = {
   materializedPeriods: RecurringAuthoringPreviewPeriod[];
 };
 
-function resolvePreviewBillingFrequency(value: string | undefined): Extract<
+type ClientPreviewBillingFrequency = BillingCycleType;
+type ContractPreviewBillingFrequency = Extract<
   BillingCycleType,
-  'monthly' | 'quarterly' | 'annually'
-> {
+  'monthly' | 'quarterly' | 'semi-annually' | 'annually'
+>;
+
+function resolveClientPreviewBillingFrequency(value: string | undefined): ClientPreviewBillingFrequency {
+  switch (value) {
+    case 'weekly':
+      return 'weekly';
+    case 'bi-weekly':
+      return 'bi-weekly';
+    case 'quarterly':
+      return 'quarterly';
+    case 'semi-annually':
+      return 'semi-annually';
+    case 'annually':
+      return 'annually';
+    case 'monthly':
+    default:
+      return 'monthly';
+  }
+}
+
+function resolveContractPreviewBillingFrequency(value: string | undefined): ContractPreviewBillingFrequency {
   switch (value) {
     case 'quarterly':
       return 'quarterly';
+    case 'semi-annually':
+      return 'semi-annually';
     case 'annually':
       return 'annually';
+    case 'weekly':
+    case 'bi-weekly':
+    case 'monthly':
     default:
       return 'monthly';
   }
@@ -66,7 +92,7 @@ function buildPreviewSourceObligation(cadenceOwner: 'client' | 'contract'): IPer
 function buildMaterializedPreviewPeriods(input: {
   cadenceOwner: 'client' | 'contract';
   billingTiming: 'arrears' | 'advance';
-  billingFrequency: Extract<BillingCycleType, 'monthly' | 'quarterly' | 'annually'>;
+  billingFrequency: string | undefined;
   t: TFunction;
 }): RecurringAuthoringPreviewPeriod[] {
   const duePosition = input.billingTiming === 'advance' ? 'advance' : 'arrears';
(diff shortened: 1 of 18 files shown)


NatalliaBukhtsik (changes requested):
resolveContractPreviewBillingFrequency silently demotes weekly to monthly
recurringAuthoringPreview.ts:24-39 — for contract cadence, weekly and bi-weekly (and any unknown value) fall back to monthly. This matches the materializer's type (materializeContractCadenceServicePeriods only supports monthly | quarterly | semi-annually | annually), but the wizard's billing-frequency dropdown exposes weekly ([packages/billing/src/constants/billing.ts:15](https://github.com/Nine-Minds/alga-psa/blob/claude/review-task-import-phase-FWoqp/packages/billing/src/constants/billing.ts#L15) — BILLING_FREQUENCY_VALUES = ['weekly', 'monthly', 'quarterly', 'annually']). Net effect: a user picking contract cadence + weekly billing sees a monthly preview with no indication that what they selected isn't honored.
12 new keys are referenced by code but absent from [server/public/locales/en/features/billing.json](https://github.com/Nine-Minds/alga-psa/blob/claude/review-task-import-phase-FWoqp/server/public/locales/en/features/billing.json)

PR #2523, opened 2026-05-18
Tickets: zoom/density controls, bulk actions, and DataTable fit/tooltip fixes

arynshimmy:
## Summary

Builds out the ticketing dashboard's list experience and hardens the shared `DataTable`:

- **Zoom/density**: granular, responsive zoom for the tickets table with readable cell wrapping; tag sizes, filter-control sizes, and the Bundled toggle now scale with the zoom level (not just spacing).
- **Bulk actions**: a floating bulk-action bar with Assign, Bundle, Move to Board, Delete, and a "More" menu (Status, Priority, Tags, Due Date). The legacy per-row three-dot actions column is removed in favor of the bar.
- **DataTable**: columns now fit the container instead of overflowing, with a "Show all" toggle; the truncation tooltip uses our custom Tooltip component instead of the native `title`.

### Zoom / density
- Granular zoom slider (10 steps) driving padding, row density, font sizes, and per-step `minColumnWidth` originally — now superseded by size-based column fit.
- Cells wrap on word boundaries instead of mid-character.
- Tag size (`sm`/`md`), filter control height/font (5 bands, every ~2 steps), and the Bundled `Switch` + label scale with the zoom level. Spacious anchors at today's sizing and scales down toward compact.

### Bulk actions
- New floating `BulkTicketActionBar` (replaces the in-header buttons); appears when tickets are selected.
- Primary buttons: **Assign**, **Bundle**, **Move to Board**, **Delete**. Secondary in a **More** dropdown: **Status**, **Priority**, **Tags**, **Due Date**.
- New dialogs: `BulkAssignTicketsDialog`, `BulkAddTagsDialog`, `BulkSetDueDateDialog`, `BulkChangeStatusDialog`, `BulkChangePriorityDialog`.
- New server actions (`bulkAssignTickets`, `bulkAddTagsToTickets`, `bulkUpdateTicketDueDate`, `bulkUpdateTicketStatus`, `bulkUpdateTicketPriority`) returning per-ticket `{ updatedIds, failed }`.
- Team assign uses the canonical flow (`assignTeamToTicket` / `removeTeamFromTicket`) so `assigned_team_id` + `team_member` resources persist; user assign clears any stale team.
- Permission-gated: action buttons hidden without `ticket:update`; bulk tag writes check `ticket:update` server-side.
- Status is disabled when the selection spans multiple boards or includes off-page tickets (select-all-matching), avoiding partial-failure paths.
- Field-edit actions (Assign/Tags/Due Date/Status/Priority) keep the selection so users can chain bulk actions; Delete/Move clear it.
- Removed the per-row "Actions" three-dot column and the now-orphaned single-ticket delete dialog chain.
- Due-date picker mounts the `Calendar` inline so it's fully visible in the dialog.
- i18n: new `bulk.*` keys added across all 10 locales.

### Selection UX
- The entire checkbox column is a click target (overlay), so clicking cell padding no longer opens the ticket.

### Shared DataTable
- Column fit: visible columns are chosen by accumulating each column's real size until the container is full, so visible columns never overflow. Adds a **Show all** / **Show less** toggle in the hidden-columns banner for horizontal scrolling.
- Truncation tooltip now uses the custom `Tooltip` (controlled open, shown only when content actually overflows) instead of the native `title` attribute; `Tooltip` gains optional `open`/`onOpenChange`/`delayDuration`.
- Removed the dead `minColumnWidth` prop now that fit is size-based.

## Test plan
- [ ] Sweep the zoom slider: padding, row density, fonts, tags, filter controls, and Bundled toggle all scale; spacious matches prior look.
- [ ] Tickets table never shows a horizontal scrollbar in the default (fitted) view; "Show all" reveals all columns with scroll, "Show less" collapses.
- [ ] Truncated cells/headers show the custom dark tooltip; non-truncated content shows none.
- [ ] Clicking anywhere in the checkbox column toggles selection (doesn't open the ticket).
- [ ] Bulk bar: Assign (user + team), Bundle, Move to Board, Delete, and More → Status/Priority/Tags/Due Date all work, with partial-failure handling.
- [ ] Team assign persists the team badge/filter; reassigning to a user clears the team.
- [ ] Status disabled for multi-board / select-all-matching selections.
- [ ] Field-edit bulk actions keep the selection; Delete/Move clear it.
- [ ] Bulk tag add denied without `ticket:update`.

### DIFF
--- a/packages/msp-composition/src/clients/MspClientTickets.tsx
+++ b/packages/msp-composition/src/clients/MspClientTickets.tsx
@@ -11,9 +11,7 @@ import { BoardPicker } from '@alga-psa/ui/components/settings/general/BoardPicke
 import CategoryPicker from '@alga-psa/tickets/components/CategoryPicker';
 import CustomSelect, { SelectOption } from '@alga-psa/ui/components/CustomSelect';
 import { getTicketsForListWithCursor } from '@alga-psa/tickets/actions/optimizedTicketActions';
-import { deleteTicket } from '@alga-psa/tickets/actions/ticketActions';
 import { XCircle } from 'lucide-react';
-import { ConfirmationDialog } from '@alga-psa/ui/components/ConfirmationDialog';
 import { useDrawer } from "@alga-psa/ui";
 import TicketDetails from '@alga-psa/tickets/components/ticket/TicketDetails';
 import { getConsolidatedTicketData } from '@alga-psa/tickets/actions/optimizedTicketActions';
@@ -72,9 +70,6 @@ const MspClientTickets: React.FC<ClientTicketsProps> = ({
   const [currentUser, setCurrentUser] = useState<IUser | null>(null);
   const [displaySettings, setDisplaySettings] = useState<TicketingDisplaySettings | null>(null);
   const ticketTagsRef = useRef<Record<string, ITag[]>>({});
-  const [ticketToDelete, setTicketToDelete] = useState<string | null>(null);
-  const [ticketToDeleteName, setTicketToDeleteName] = useState<string | null>(null);
-  const [deleteError, setDeleteError] = useState<string | null>(null);
   const [isQuickAddTicketOpen, setIsQuickAddTicketOpen] = useState(false);
 
   const { openDrawer } = useDrawer();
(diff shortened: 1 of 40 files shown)


NatalliaBukhtsik (changes requested):
Permission check ordered after data read — packages/tickets/src/actions/ticketActions.ts:1808-1820 bulkAddTagsToTickets calls findTagsByEntityIds(uniqueIds, 'ticket') before hasPermission('ticket','update') at :1812+. A caller without update can still enumerate existing tags. Reorder: permission first, then read.
Bulk loops are N round‑trips — every bulk* action serially calls updateTicketWithCache, which opens its own transaction and re‑runs hasPermission per ticket (optimizedTicketActions.ts:1912). "Select all matching" → hundreds of permission lookups + hundreds of transactions. Hoist the permission check; do one transaction per action (or WHERE id = ANY($1)).
4 bulk actions lack an entry permission check — bulkAssignTickets, bulkUpdateTicketDueDate, bulkUpdateTicketStatus, bulkUpdateTicketPriority rely solely on per‑ticket checks inside the loop. Add hasPermission(user, 'ticket', 'update', knex) at the top of each, matching bulkAddTagsToTickets:1812.
BulkSetDueDateDialog timezone bug — BulkSetDueDateDialog.tsx:48 does date.toISOString(), sending UTC midnight. Users west of UTC will see the date shift back a day. Use a date‑only string or apply the user's TZ.
selectedTicketsSharedBoardId returns null whenever any selected ID isn't on the current page — TicketingDashboard.tsx:929-939. The Status action then silently disables on the common paginate‑then‑select pattern and the fallback message ("on different boards") is misleading. Either fetch board IDs for off‑page rows or change the disabled tooltip.

NatalliaBukhtsik (approved):

PR #2536, opened 2026-05-20
Overhaul project task list view: styling, columns, zoom, inline editing

arynshimmy:
## Summary

Brings the project detail **List view** (`TaskListView`) up to parity with the restyled DataTable and the tickets table, and makes it directly editable.

### Styling & structure
- Restyled to match the shared DataTable: rounded-xl card, tinted header, CSS-variable color tokens (dark-mode aware), thin divide lines + hover, vertical column separators, 12px/13px type.
- Columns are now **data-driven** — header, colgroup, and body cells all render from one ordered column config, so they always stay aligned.

### Per-phase status fix
- Added `getProjectStatusesByPhase` so the List view buckets each phase's tasks by **that phase's own** effective status mappings (mapping IDs are phase-specific). Fixes phases that showed a task count but empty status groups.

### Columns
- Added **Status**, **Priority**, and **Type** columns; reordered all columns by importance/relevance.
- Greedy real-width responsive fit (far fewer columns hidden) with a **Show all / Show less** toggle, plus **draggable column resizing** persisted per project.

### Zoom / density
- Added a tickets-style **ViewDensityControl** next to the kanban/list toggle (server-persisted via user preferences). Scales row padding, text, in-cell pickers/tags/avatars, and the column fit — so the number of visible columns changes with zoom.

### Inline editing
- Edit **status, priority, type, and due date** via the same pickers used in the task edit form; edit **estimated/actual hours** by clicking to type. All persist through a single partial-update handler.

### Shared component fixes
- `CustomSelect` / `DatePicker` / `TaskStatusSelect` / `TaskTypeSelector` / `PrioritySelect` now truncate the trigger value with an ellipsis instead of hard-clipping in narrow columns.

### i18n
- New keys (`projectList.columns.priority`/`taskType`, `dataTable.*`) added across all locales; real locales translated, `xx`/`yy` pseudo-locales regenerated.

## Testing
- `T001` lang-pack key-parity tests pass; `scripts/validate-translations.cjs` reports 0 errors.
- The pre-existing `T002` "zero-string components" test failure (`ProjectPage.tsx`/`TaskEdit.tsx` already contain `useTranslation` on `main`) is unrelated to this branch.
- UI changes verified manually in the running app.

## Docs
- A companion docs page (`6.7 View and Edit Project Tasks in the List View`) was drafted in the nm-store repo (separate repo, not part of this PR).


NatalliaBukhtsik (changes requested):
N+1 transactions in getProjectStatusesByPhase + silently swallowed errors - worsens the preexisting pattern
Unnecessary getTaskChecklistItems refetch on every inline edit
Column widths stored only in localStorage — TaskListView.tsx:243-264. Other prefs in this view (density, kanban zoom, panel visibility) are server-persisted via prefs

NatalliaBukhtsik (approved):

PR #2572, opened 2026-05-27
Bulk-apply tags to selected project tasks

arynshimmy:
## Summary
- Adds a **Tags** entry to the project-task multi-select toolbar, mirroring the ticket bulk-tag flow: pick tasks, open the dialog, type one or more tags, and they're applied to every selected task (tags already on a task are skipped).
- New `bulkAddTagsToTasks` server action: tenant-scoped `project_tasks` existence filter (so unknown UUIDs cannot mint orphan `tag_mappings`), dedupes IDs/tag texts, runs per-task in its own transaction, reports partial failures with task names so the UI narrows selection to just the failed rows.
- Hardens the shared tag helper used by both the new task bulk path and the existing ticket bulk path:
  - `createTagsForEntityWithTransaction` now mirrors `createTag`'s `tag:create` gate — minting a brand-new tag definition requires `tag:create`; permission errors re-throw out of the per-tag catch so the caller can attribute the failure.
  - New exported helpers `captureEntityTagTextSnapshot` + `publishEntityTagBulkUpdate` let bulk callers emit the per-entity update event (`PROJECT_TASK_UPDATED` / `TICKET_UPDATED`) that the helper itself deliberately suppresses. The webhook-emission contract test still passes — its assertion is scoped to the helper's own source.
- Bulk task action compares the helper's returned tag list against the requested texts and surfaces dropped tags in `failed` (length-overflow, unique-conflict, etc. no longer silently report success).
- i18n: adds `bulkActions.tags`, `bulk.tags.*`, and `projectDetail.bulkTags*` (success_one/_other, partial, failure) to `en` plus all 8 non-English projects locales (`fr/es/de/nl/it/pl/xx/yy`) so the projects i18n audit's per-locale key-count parity stays green.

## Test plan
- [ ] Open a project with several tasks across phases and statuses; multi-select 3+ rows and click the new **Tags** action.
- [ ] Submit a mix of new + already-present tags — confirm the new ones appear inline in each row without page reload and the already-present tag is not duplicated.
- [ ] Submit a tag that exceeds 50 chars in the mix — the task lands in the failed list with a message naming the dropped tag; valid tags on other tasks still apply.
- [ ] As a user with `project:update` but without `tag:create`, try to apply a brand-new tag in bulk — every task lands in failed with a permission message; applying an existing tag works.
- [ ] Run with one of the selected task IDs deliberately invalid (force a stale selection) — that task lands in failed with "Task not found", others succeed.
- [ ] Verify the existing single-task `TagManager` (inline in the list row and inside the Edit Task dialog) still works unchanged.
- [ ] Confirm `npm run typecheck` passes in `packages/projects`, `packages/tags`, and `packages/tickets`.
- [ ] Confirm the tag webhook-emission contract test passes (`packages/tags` → `tagActions.webhookEmission.contract.test.ts`).
- [ ] Confirm the projects i18n-audit key-parity tests pass for all 8 non-English locales.

## Docs
- Companion doc edit in nm-store: `67-using-the-project-list-view.md` gains a new **Selecting Tasks for Bulk Actions** section describing Move / Assign / Tags / Delete and the partial-failure narrowing behavior.

### DIFF
--- a/packages/projects/src/actions/projectTaskActions.ts
+++ b/packages/projects/src/actions/projectTaskActions.ts
@@ -21,9 +21,16 @@ import type {
   ITaskType,
   IUser,
   IUserWithRoles,
+  PendingTag,
   ProjectStatus,
 } from '@alga-psa/types';
-import { findTagsByEntityIds } from '@alga-psa/tags/actions';
+import { revalidatePath } from 'next/cache';
+import {
+  captureEntityTagTextSnapshot,
+  createTagsForEntityWithTransaction,
+  findTagsByEntityIds,
+  publishEntityTagBulkUpdate,
+} from '@alga-psa/tags/actions';
 import { withAuth } from '@alga-psa/auth';
 import { hasPermission } from '@alga-psa/auth/rbac';
 import { validateArray, validateData } from '@alga-psa/validation';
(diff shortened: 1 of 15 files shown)


NatalliaBukhtsik (changes requested):
Per-task transaction is N transactions for N tasks.
bulkAddTagsToTasks opens a new withTransaction inside the per-task loop. Reasonable for isolation (one failure doesn't roll back the others) and lets the bulk-event publish stay scoped, but selecting 200 tasks issues 200 round-trip transactions. 

Case-sensitivity is inconsistent within the bulk action.
normalizedTexts dedupes by exact trimmed text (new Set(tagTexts.map(t => t.trim()).filter(...))), but the existing-tag check downstream uses .toLowerCase() (alreadyOnTask.has(t.toLowerCase())).

NatalliaBukhtsik (approved):

PR #2619, opened 2026-06-03
Add per-user hide status columns to project Kanban view

arynshimmy:
## Summary

Adds the ability to **optionally hide one or more status columns** in the project Kanban view. This is a purely visual, per-user convenience — it does **not** change the admin-configured status visibility (`is_visible`) for the project, so hiding a column only affects *your own* board. The main use case is decluttering a wide board to make it easier to drag tasks across far-apart columns.

## What's included

- **Quick-hide button per column** — an eye-off icon in each Kanban column header (next to the task-count badge) hides that column in one click.
- **Columns popover in the toolbar** — a new eye-off toolbar button (alongside the existing zoom / sticky-names / pin controls) opens a popover listing every visible status, each with an eye / eye-off toggle, plus a **Show all** button to restore everything at once. The toolbar button shows a small count badge while any columns are hidden.
- **Persistence** — hidden columns are stored as a new batched user preference, `project_kanban_hidden_statuses:<projectId>`, via the existing `useUserPreferencesBatch` hook (localStorage + server, debounced). It is scoped per project since each project has its own statuses.
- **Consistency** — both the Kanban board and the sticky status-name strip respect the hidden set. Stale IDs (e.g. a status later deleted or made invisible by an admin) are guarded against so they don't affect the count or display.

## Design notes

- Hiding is **layered on top of** the admin `is_visible` filter rather than replacing it — `displayedKanbanStatuses = visibleKanbanStatuses − hiddenStatusIds`.
- The feature is always recoverable: even if every column is hidden, the toolbar popover and **Show all** remain available.

## Files changed

- `packages/projects/src/components/ProjectDetail.tsx` — new preference key/entry; `displayedKanbanStatuses` derived state; `toggleKanbanStatusHidden` / `showAllKanbanStatuses` helpers; the Columns toolbar popover; wired the displayed set into the board and sticky strip.
- `packages/projects/src/components/KanbanBoard.tsx` — pass-through `onHideColumn` prop.
- `packages/projects/src/components/StatusColumn.tsx` — `onHideColumn` prop and the per-column hide button.

## Testing

- Type-checks cleanly (`tsc --noEmit` on the `projects` package).
- Manually verified in the running app: hiding via both the column button and the popover, the count badge, **Show all**, persistence across reloads, and correct behavior of the sticky status strip.


NatalliaBukhtsik (changes requested):
Doesn't account for per-phase custom statuses, will (skip?) phases where statuses have been customized even though hidden status exists
No update/reconsiliation if someone customizes phase after hidden statuses have been recorded
No cleanup of preferences on status/project/user deletion
Hiding every column leaves a blank board — recoverable only via the toolbar popover (which lists visibleKanbanStatuses), but the per-column eye-off buttons disappear, so it may feel like a dead end.

PR #2656, opened 2026-06-09
feat(appliance): daily connected-license check-in via Temporal

RobertAtNineMinds:
## What

Adds the **daily connected-license check-in** — the missing half of the appliance license renewal loop — as a Temporal schedule in the appliance worker.

The alga-license `/check-in` endpoint and the appliance-stored `appliance_credential` + `check_in_url` already existed, but **nothing ever called check-in**. A connected appliance therefore rode its ~31-day token and silently dropped to Essentials at expiry with no renewal, and a cancelled/revoked subscription never propagated until that expiry.

## How

- **`appliance-check-in-activities.ts`** — reads the `license_state` singleton; no-ops on SaaS/cloud (no row) and on non-connected installs (essentials/airgap/CE/trial); else POSTs `{appliance_credential}` to `check_in_url`. On `{status:'ok',jwt}` it stores the (rolled-forward) token + `last_checkin_at`; on `{status:'revoked'}` it **honors grace** (records it, leaves the token to expire naturally — a transient/erroneous `revoked` must not instantly down a paying customer). Transient network/5xx throw → Temporal retries; deterministic 4xx is recorded, not spun on.
- **`appliance-check-in-workflow.ts`** + **`setupSchedules.ts`** — daily schedule on the `tenant-workflows` queue (the one the appliance worker polls) with `overlap: SKIP`, plus a best-effort boot trigger so a box that was powered off renews promptly. Mirrors the existing `premium-trial-expiry` daily-maintenance pattern.
- Registered in both activity/workflow barrels; 2 registration-test assertions added.
- **Doc fix:** `license_state.appliance_credential` is stored in **plaintext** (the check-in caller needs it), not a SHA-256 hash — corrected the misleading comment.

Decision (with @robertisaacs): on revocation we **honor grace** rather than hard-cut.

## Validation

Validated **end-to-end** against prod from the test appliance: the real appliance credential POSTed to `/check-in` returned `status=ok, tier=pro, aud=<tenant-bound>` — exercising the entire path the worker drives. Plus `tsc --noEmit` clean and the registration-test assertions.

## Two related bugs found while wiring this (both pre-existing, latent)

1. **Gateway routed a dead path.** The istio VirtualService matched `exact: /checkIn`, but Fastify serves `/check-in` and `register.ts` hands every appliance a `…/check-in` URL — so `POST /check-in` 404'd for everyone. **Fixed + applied to prod separately** (`nm-kube-config/alga-license/istio-gateway.yaml`; verified 401 not 404). ⚠️ That nm-kube-config file is currently *untracked* in its repo — it should be committed there for the record.
2. **temporal-worker image `npm ci` ERESOLVE.** `Dockerfile.production` didn't pass `legacy-peer-deps`, so `npm ci` died on `nodemailer@^8` vs `next-auth` peerOptional `^7`. **Fixed in this PR** (`NPM_CONFIG_LEGACY_PEER_DEPS=true` in the base stage).

## Known, out of scope here

The temporal-worker **image still can't build from `main`** after the `npm ci` fix: `Dockerfile.production` doesn't install the modularized `packages/*` workspace (transitive `@alga-psa/user-composition` 404s). This is the known modularization blocker (≈3 issues: workspace install, nx cycle, missing type) and also affects `workflowWorker` — tracked as a **separate remediation**. Until that image rebuilds cleanly, this feature can't be *deployed* to a worker, but the feature + gateway fix are validated above.

### DIFF
--- a/ee/temporal-workflows/Dockerfile.production
+++ b/ee/temporal-workflows/Dockerfile.production
@@ -7,6 +7,13 @@ RUN apt-get update && apt-get install -y \
 
 WORKDIR /app
 
+# The repo root .npmrc sets legacy-peer-deps=true, but this build copies only
+# package*.json (not .npmrc) before running npm, so npm would fall back to
+# strict peer resolution and fail ERESOLVE (e.g. nodemailer@^8 vs next-auth's
+# peerOptional nodemailer@^7). Set it via env so every npm step (ci + install,
+# in any WORKDIR) honors it — matching the repo .npmrc and the EE image build.
+ENV NPM_CONFIG_LEGACY_PEER_DEPS=true
+
 # Development stage - build everything
 FROM base AS development
 
--- a/ee/temporal-workflows/src/__tests__/worker-registration.test.ts
+++ b/ee/temporal-workflows/src/__tests__/worker-registration.test.ts
@@ -22,4 +22,12 @@ describe('temporal worker registration', () => {
   it('exports proactive NinjaOne token refresh activity from activities index', () => {
     expect(activities.proactiveNinjaOneTokenRefreshActivity).toBeDefined();
   });
+
+  it('exports applianceCheckInWorkflow from workflow index', () => {
+    expect(workflows.applianceCheckInWorkflow).toBeDefined();
+  });
+
+  it('exports applianceLicenseCheckInActivity from activities index', () => {
+    expect(activities.applianceLicenseCheckInActivity).toBeDefined();
+  });
 });
--- a/ee/temporal-workflows/src/activities/appliance-check-in-activities.ts
+++ b/ee/temporal-workflows/src/activities/appliance-check-in-activities.ts
@@ -1,0 +1,136 @@
+/**
+ * Appliance license check-in activity.
+ *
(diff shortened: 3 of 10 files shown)


RobertAtNineMinds:
Lead for the separate temporal-worker image remediation (from memory `appliance-image-build-blockers` + this run):

The workspace-aware Dockerfile fix already exists in **`ee/temporal-workflows/Dockerfile`** (commit `d72eab3b`: trimmed workspaces glob + single `npm install --legacy-peer-deps` + `COPY --from=development /app /app`, verified via `docker build --target production`). But the Argo `temporal-worker-build` workflow (nm-kube-config) still builds **`--file ee/temporal-workflows/Dockerfile.production`** — the stale per-subdir variant that only copies `shared/` and 404s on transitive `@alga-psa/user-composition`.

So the likely fix for the worker image is **repointing the build to `--file ee/temporal-workflows/Dockerfile --target production`** (in nm-kube-config's temporal-worker-build-workflow.yaml), not patching Dockerfile.production. The `legacy-peer-deps` change in this PR is then a harmless no-op for the real build path (it targets the stale file). Worth confirming with one build before deleting Dockerfile.production.

PR #2678, opened 2026-06-11
fix(licensing): fresh installs aren't 'Trial Expired' + immediate session tier refresh

RobertAtNineMinds:
Two self-host trial UX fixes found while assessing whether the essentials→trial journey is adequately tested (full VM smoke of that journey happening next, will report results there).

## 1. `trial_available` state for fresh installs

An EE install that never started its trial resolved to `trial_expired` (license-state case 4), so a **day-one essentials box** showed:
- the app-wide warning banner *“Enterprise trial has expired. The install is now running Essentials features.”*
- License page status *“Trial Expired — Essentials”*

Now: new `trial_available` state → info banner *“Running Essentials features. Start a free 30-day Enterprise trial to unlock all features.”* + label *“Essentials — Enterprise Trial Available”*. `trial_expired` (and its warning banner) is reserved for an actually-elapsed trial.

`LicenseStatus.state` types flow through from `LicenseStateKind`, and the only state consumers are the License page label map and `LicenseBanner` — both updated. `canStartTrial` already keyed off `trialUsed`/`licensed`, unchanged.

## 2. Immediate session tier refresh on license changes

`user.effectiveTier` is stamped into the NextAuth JWT and only re-resolved every **5 minutes** (`PLAN_CHECK_INTERVAL`), so starting a trial (or activating/connecting a license) left tier-gated features stale for up to 5 minutes or until re-login. The License page now calls `useSession().update()` after every successful license-state mutation — that's the existing `trigger === 'update'` force-refresh path in the jwt callback. (`/msp` is wrapped in `SessionProvider` via `MspLayoutClient`.)

## Tests

`packages/licensing`: 22/22 vitest, `tsc --noEmit` clean. The fresh-install assertion updated to expect `trial_available`; the elapsed-trial case still asserts `trial_expired`.

### DIFF
--- a/packages/licensing/src/lib/license-state.test.ts
+++ b/packages/licensing/src/lib/license-state.test.ts
@@ -87,7 +87,9 @@ describe('resolveSelfHostTier', () => {
     // with no trial and no license.
     const result = resolveSelfHostTier(makeRow({ edition_choice: 'ee' }));
     expect(result?.tier).toBe('essentials');
-    expect(result?.state).toBe('trial_expired');
+    // Fresh ee install with no trial used: trial_available, NOT trial_expired —
+    // the expired state (and its banner) is reserved for an actually-elapsed trial.
+    expect(result?.state).toBe('trial_available');
   });
 
   it('license takes precedence over an active trial', () => {
--- a/packages/licensing/src/lib/license-state.ts
+++ b/packages/licensing/src/lib/license-state.ts
@@ -34,6 +34,7 @@ export interface LicenseStateRow {
 /** Derived licensing state for an install. */
 export type LicenseStateKind =
   | 'ce'            // Community Edition chosen — always essentials
+  | 'trial_available' // EE install, no license, trial not yet used — essentials, trial can be started
   | 'trial'         // Enterprise trial active
   | 'trial_expired' // Trial window elapsed, no license
   | 'licensed'      // Valid unexpired license present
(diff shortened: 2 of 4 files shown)


RobertAtNineMinds:
**Live-verified on the appliance VM** (fresh essentials install-code install from the stable channel, then upgraded to this PR's image `932d3756`):

- **Pre-fix (stable `9468067e`), reproduced:** fresh never-trialed install showed the app-wide warning banner *"Enterprise trial has expired"* + License page *"Trial Expired — Essentials"*.
- **Fix 1 (`trial_available`):** same box on `932d3756` shows the info banner *"Running Essentials features. Start a free 30-day Enterprise trial…"* + *"Essentials — Enterprise Trial Available"*.
- **Fix 2 (session refresh):** primed the JWT's `last_plan_check` seconds before clicking Start 30-Day Trial (so the 5-min interval couldn't explain a flip) → `/api/auth/session` returned `effectiveTier: premium` **seconds after the click**.
- **Full journey also verified:** real button → state `trial`/premium/30 days; trial card hidden once used; backdating `trial_started_at` −31d → read-time downgrade to essentials, legitimate "trial has expired" warning, label "Trial Expired — Essentials", one-trial-per-install guard holds; session catches up via the 5-min interval.

Side-finding (separate issue-worthy): `NEXTAUTH_URL` defaults to `https://alga.local` → `__Secure-`/`__Host-` cookies → **real browsers cannot keep a session on `http://<ip>:3000`** (day-one default access). The smoke required pointing `NEXTAUTH_URL` at the IP.

PR #2682, opened 2026-06-11
Fix QBO smoke-run findings (alga0001975–alga0001982)

RobertAtNineMinds:
Fixes the eight tickets filed from the 2026-06-11 QBO Online integration live smoke run.

## Sync engine (alga0001975, alga0001976, alga0001979)
- **alga0001975** — Scheduled-export **validation failures now file the exception immediately** instead of after the 5-attempt cap (~75 min of silence). `executeBatch` throws a typed `ACCOUNTING_EXPORT_VALIDATION_FAILED` carrying the open validation errors; transient errors keep retry-then-alert; a later successful export resolves the exception.
- **alga0001976** — A failed scheduled batch can no longer **wedge the duplicate-selection guard**: the drain auto-cancels its own engine-owned batch on failure, self-heals already-wedged tenants (cancel stale scheduled batch + retry once), and never touches manual batches. The batch detail panel gains a guarded **Cancel** action (`cancelAccountingExportBatch`) so operators can clear wedged batches without SQL.
- **alga0001979** — Pushed payments that QBO silently books as **unapplied customer credit** (`UnappliedAmt > 0`, application line dropped) now file an `accounting_sync_unmapped_payment` exception (`pushed_payment_unapplied`). Op stays done — the money landed.

## Credit flows (alga0001978, alga0001980, alga0001981)
- **alga0001978** — QBO's **"Automatically apply credits"** (default ON) racing Alga-driven application: health panel reads the Preferences API and warns with the exact settings path; the `apply_credit` applier reads the CreditMemo's remaining balance before pushing the zero-dollar linking Payment and files a conflict exception instead of double-applying (self-heals when the credit is freed).
- **alga0001980** — Voiding a credit note now **claws back its issued pool credit**. Root cause: the claw-back queried `credit_issuance` while credit notes write `credit_issuance_from_negative_invoice`, so it silently no-opped. Both the consumed-credit guard and the claw-back match both types, and the claw-back writes an auditable `credit_adjustment` transaction. No `CREDIT_NOTE_VOIDED` consumer exists, so the synchronous claw-back is the fix.
- **alga0001981** — `CreditApplicationUI` is no longer orphaned: the **finalized-invoice preview panel** gains an Apply Credit action so operators can apply chosen/partial credit to a chosen invoice (and deliberately trigger the `apply_credit` sync producer).

## Defaults & RBAC (alga0001977, alga0001982)
- **alga0001977** — Onboarding wizard **auto-sync defaults OFF** per plan F004; the go-live switch stays for explicit opt-in, with copy explaining the consequence and where to enable later. Contract test updated to the intended default.
- **alga0001982** — Seeds the missing **`financial` resource permissions** (create/read/update/delete/transfer) in onboarding + dev seeds and backfills existing tenants via migration, granting Admin and Finance MSP roles. Fresh-tenant Admins can now use every documented `/api/v1/financial` endpoint. Option 1 (seed) chosen over remapping controllers since the controller spans transactions/payment methods/tax/reports.

## Test notes
- vitest 4 invokes constructor mocks with real `new` semantics; the arrow-function factory mocks in `accountingSyncCycleService.test.ts` and `syncProducers.test.ts` were failing **at baseline** and are now constructible function expressions.
- 146 tests across 14 touched suites pass; `packages/billing`, `packages/types`, `packages/integrations` typecheck clean.
- Pre-existing unrelated failure left as-is: `InvoicePreviewPanel.test.tsx` template-resolution tests mock an API the component no longer calls (flagged separately).


RobertAtNineMinds:
Closing/reopening to retrigger CI — the GitHub webhooks incident dropped the synchronize events for the last two pushes.

PR #2758, opened 2026-06-23
Restrict Tactical RMM raw command workflows with RBAC

robertisaacs:
### Motivation
- Prevent workflows from abusing stored Tactical RMM credentials to run arbitrary OS commands by requiring an explicit RMM permission on raw command execution. 
- Preserve existing administrator workflows while blocking non-RMM workflow publishers from escalating to endpoint command execution.

### Description
- Add an RBAC check to the `tacticalrmm.agents.run_command` handler by introducing `RMM_COMMAND_PERMISSION = { resource: 'rmm', action: 'execute_command' }` and calling `requireRmmCommandPermission(ctx)` before invoking the vendor client. (ee/packages/workflows/src/runtime/actions/registerTacticalRmmWorkflowActions.ts)
- Implement the permission check using existing workflow helpers via `withTenantTransaction` + `requirePermission` so the workflow actor is resolved from the run context before the call. (shared workflow businessOperations import usage)
- Add a migration that creates the new MSP-only `rmm:execute_command` permission and grants it to existing MSP Admin roles, and add the permission to PSA onboarding seeds so new tenants receive it. (server/migrations/20260623090000_add_rmm_execute_command_permission.cjs, ee/server/seeds/onboarding/psa/02_permissions.cjs)
- Update unit tests to cover both allowed and denied execution paths and ensure the Tactical RMM client is not called when permission is missing. (ee/packages/workflows/src/runtime/actions/__tests__/tacticalRmmWorkflowActions.handlers.test.ts)

### Testing
- Ran the workflows unit test suite for the modified module: `cd ee/packages/workflows && npx vitest run src/runtime/actions/__tests__/tacticalRmmWorkflowActions.handlers.test.ts src/runtime/actions/__tests__/tacticalRmmWorkflowClient.test.ts` (run inside the monorepo with a local stubbed config during verification). Result: tests in these files passed (12 tests passed when executed in the workflows package test environment).
- Also executed the client tests with a custom Vitest config used during development: `npx vitest run --config /tmp/workflows-vitest.config.mjs ...` which exercised the new behavior and helped validate the permission gating (tests for the two affected test files passed in that environment).
- Type-check/build was not completed in this ephemeral environment due to missing ambient test/dev deps (React/Next/testing-library types) so a full repo `tsc` was not validated here; migration and seed changes are included and should be applied in CI/DB upgrade workflows.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a39ed9379188330a4f0ccf95d349c3d)


RobertAtNineMinds:
Re: the AI review — the "blocking" items are based on a truncated diff; the migration is complete and correct.

**`down()` is not truncated.** Full rollback (`server/migrations/20260623090000_add_rmm_execute_command_permission.cjs`):
```js
exports.down = async function down(knex) {
  const permissions = await knex('permissions')
    .where({ resource: 'rmm', action: 'execute_command' })
    .select('tenant', 'permission_id');

  for (const { tenant, permission_id } of permissions) {
    await knex('role_permissions').where({ tenant, permission_id }).del();
    await knex('permissions').where({ tenant, permission_id }).del();
  }
};
```

- **Tenant scoping:** the rollback intentionally targets the permission across all tenants (mirroring the per-tenant `up` loop), and every `.del()` is scoped by `(tenant, permission_id)` — so it's tenant-safe on the distributed table.
- **Duplicates:** `up` does a `.where(...).first()` existence check before inserting both the permission and each `role_permissions` row, so it's idempotent and won't create duplicates.
- **Citus validity:** the **Combined migrations on single-node Citus** CI job (which actually runs this migration) is green.

The "Should Fix"/"Suggestions" items (shared error helper, RBAC-before-integration-lookup ordering, docs) are reasonable stylistic notes but are non-blocking and don't affect correctness. Proceeding to merge once CI completes.

---

PR #2829, opened 2026-07-01
Port outbound SMTP TLS controls and connection test to EE ManagedEmailSettings

RobertAtNineMinds:
## Summary

Follow-up to #2823 for ticket alga0002046 ("Appliance outbound SMTP fails with misleading 'Email service is disabled or not configured'").

#2823 added no-auth relay support, TLS controls, and a Test Connection card for outbound SMTP — but the UI half landed only in the OSS `EmailSettings.tsx`. The appliance runs enterprise edition, which renders the EE `ManagedEmailSettings.tsx` instead (`@ee` alias), so none of those controls reached the reporter. This PR ports them to the EE component:

- **Connection Security toggles** on the outbound SMTP card — implicit TLS (`secure`), Require STARTTLS (`requireTLS`), and Verify server certificate (`rejectUnauthorized`, with self-signed-relay hint) — wired to the same config fields the backend already reads. Unblocks self-signed LAN relays.
- **Test Connection section** on the SMTP card — persists current edits (masked `***` password resolves to the stored secret server-side), runs the existing `testOutboundEmail` action, and surfaces the real provider error instead of a generic message; optional recipient field sends a test message.
- **No-auth hint** under username/password, matching the OSS form.
- Export `testOutboundEmail` from the integrations actions barrel (previously only in the email-actions sub-barrel).
- Add the 13 new `msp/email-providers` i18n keys across all 10 locales, reusing the translations #2823 added to `msp/admin`; regenerated `xx`/`yy` pseudo-locales (also picks up stale `neverSeenWarning` `{{domain}}` drift in `msp/settings`).
- Repair `ManagedEmailSettings.actions.test.tsx`, which was failing on main (`useTier` outside `TierProvider`, empty i18n resources; CI excludes the `sebastian-ee` unit suite so it went unnoticed) and add 4 tests covering the new SMTP connection test and TLS toggle flows.

## Verification

- `nx typecheck sebastian-ee` and `nx typecheck @alga-psa/integrations` clean
- EE unit suite `ManagedEmailSettings.actions.test.tsx` 6/6 passing (2 repaired + 4 new)
- Integrations email-actions tests 8/8 passing
- `scripts/validate-translations.cjs` and `scripts/find-missing-i18n-keys.cjs` both pass

Still open after merge: live UI smoke test of the Outbound Email tab on the appliance/dev.

https://claude.ai/code/session_01MLncqkn5oWX8eetRuKN6nY

---
_Generated by [Claude Code](https://claude.ai/code/session_01MLncqkn5oWX8eetRuKN6nY)_

### DIFF
--- a/ee/server/src/__tests__/unit/ManagedEmailSettings.actions.test.tsx
+++ b/ee/server/src/__tests__/unit/ManagedEmailSettings.actions.test.tsx
@@ -1,7 +1,9 @@
 // @vitest-environment jsdom
 import React from 'react';
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
-import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { beforeAll, beforeEach, describe, expect, it, vi } from 'vitest';
+import i18n from 'i18next';
+import emailProvidersEn from 'server/public/locales/en/msp/email-providers.json';
 import ManagedEmailSettings from '@ee/components/settings/email/ManagedEmailSettings';
 
 const {
@@ -12,6 +14,7 @@ const {
   getEmailSettingsMock,
   updateEmailSettingsMock,
   getEmailProvidersMock,
+  testOutboundEmailMock,
   toastSuccessMock,
   toastErrorMock,
 } = vi.hoisted(() => ({
@@ -22,6 +25,7 @@ const {
   getEmailSettingsMock: vi.fn(),
   updateEmailSettingsMock: vi.fn(),
   getEmailProvidersMock: vi.fn(),
+  testOutboundEmailMock: vi.fn(),
   toastSuccessMock: vi.fn(),
   toastErrorMock: vi.fn(),
 }));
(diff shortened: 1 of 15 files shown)


RobertAtNineMinds:
Verified the review's blocking questions against the implementation:

**1. Permission check on `testOutboundEmail`** — The action (added in #2823, `packages/integrations/src/actions/email-actions/emailSettingsActions.ts:215`) is wrapped in `withAuth`, which requires an authenticated user and rejects otherwise. It does not do an additional RBAC `hasPermission` check — but that matches the pre-existing posture of `getEmailSettings`/`updateEmailSettings` in the same file (the latter being strictly more sensitive than a connection test). This PR only re-exports the action from the integrations barrel; adding RBAC to just this one action would be inconsistent. A permissions pass over all three email-settings actions is worth doing as its own change.

**2. Tenant context** — `withAuth` derives tenant from the authenticated user (`user.tenant`) and runs the action inside `runWithTenant` (AsyncLocalStorage), so `createTenantKnex()`/`TenantEmailService.testConnection(tenant, …)` are scoped server-side. The only client-supplied argument is the optional recipient address; no client-provided tenant or settings object is trusted.

**Should-fix items:**
- Unique IDs: all rendered controls have them — `smtp-secure`, `smtp-require-tls`, `smtp-reject-unauthorized`, `test-recipient`, `test-outbound-email`, `save-smtp-settings`.
- Error handling: `TenantEmailService.testConnection` returns a structured `{ success, message?, error? }` with the real init/verify error (`error.message` from the provider), and the action propagates it; nothing is logged-and-swallowed.

**Suggestions:** The TLS controls and Test Connection section render inside the `outboundProvider === 'smtp'` card, so they're structurally absent for the managed/Resend provider. Skipping the import reformat and the extra visibility test for now to avoid re-triggering the full CI pipeline mid-run.

---

PR #2830, opened 2026-07-01
Fix missing interaction permissions on appliance/CE installs (alga0002048)

RobertAtNineMinds:
## Summary

Fixes ticket **alga0002048**: on appliance/CE installs the MSP Contacts page and the client-portal invite flow fail with `Permission denied: Cannot read interaction types`.

**Root cause:** the PSA onboarding seeds (`ee/server/seeds/onboarding/psa/`) that provision the first tenant during appliance bootstrap (and hosted tenants via the Temporal `run_onboarding_seeds` activity) never defined any `interaction` permissions, so the default Admin — and every other MSP role — had no `interaction:*` grants. The Contacts page and portal-invite flow read interaction types gated on `interaction:read`, so the ABAC check threw.

The earlier backfill migration (`20260622120000_add_interaction_permissions`) never repaired appliances because the appliance bootstrap runs `knex migrate:latest` against an **empty** database before the tenant is created and seeded — the backfill was recorded as applied without inserting anything.

## Changes

- **`ee/server/seeds/onboarding/psa/02_permissions.cjs`** — add the four `interaction` permission defs (msp-only), matching the dev seed. MSP Admin picks them up automatically via its all-msp-permissions grant.
- **`ee/server/seeds/onboarding/psa/03_role_permissions.cjs`** — grant interaction perms to the other standard MSP roles, mirroring the dev-seed/backfill distribution: Finance create/read/update/delete; Manager, Project Manager, Dispatcher, Technician create/read/update.
- **`server/migrations/20260701120000_readd_interaction_permissions_for_seeded_tenants.cjs`** — re-run the idempotent backfill for tenants provisioned after 20260622120000 was applied (existing appliance installs, plus any hosted tenants created before the seed fix). `down` is a deliberate no-op so it cannot strip grants created by the earlier migration or the seeds.

## Verification

Ran an end-to-end check against an ephemeral Postgres 16 (18/18 assertions passed):

- A tenant seeded with the **pre-fix** onboarding seeds reproduces the bug exactly (zero `interaction` permission rows).
- The new migration backfills all four permission rows and the per-role grants (Admin/Finance 4, Manager/PM/Dispatcher/Technician 3), and is idempotent on re-run.
- A tenant seeded with the **fixed** seeds gets all grants directly, with Admin holding every MSP permission including `interaction:read` — the exact permission the Contacts page checks.

https://claude.ai/code/session_01MLwbokjYjoeXK9vAi8Mqie

---
_Generated by [Claude Code](https://claude.ai/code/session_01MLwbokjYjoeXK9vAi8Mqie)_


RobertAtNineMinds:
Re: the AI review's blocking item — the migration file is complete on the branch; the review just saw a truncated diff. The portion after the quoted cutoff is [here](``https://github.com/Nine-Minds/alga-psa/blob/claude/alga0002048-t3f9x4/server/migrations/20260701120000_readd_interaction_permissions_for_seeded_tenants.cjs#L109-L128).`` Addressing each point:

- **Tenant scoping:** every query passes `tenant` explicitly — the roles lookup is `.where({ tenant, msp: true }).whereIn('role_name', ...)`, and `ensurePermission`/`assignPermission` both filter on `tenant` in every select/insert/update. There are no global role lookups; the loop runs entirely per-tenant. This mirrors the already-shipped `20260622120000_add_interaction_permissions` migration (same helpers, same Citus-compatible raw-knex pattern).
- **Idempotency:** verified empirically against Postgres 16 — a tenant seeded with the pre-fix onboarding seeds has zero `interaction` rows, the migration backfills 4 permission rows + 17 role grants, and a second `up()` run changes nothing (row counts identical). `down` is a deliberate no-op so it can't strip grants created by the earlier migration or the seeds.
- **Resource name consistency:** the gated actions in `packages/clients/src/actions/interactionActions.ts` and `interactionTypeActions.ts` all call `assertMspPermission(user, 'interaction', ...)` — resource `'interaction'`, matching the seeded rows. No `interaction_type` resource exists in the permission checks.

---

PR #2832, opened 2026-07-02
feat(appliance): surface "update available" in status UI and document the upgrade path

RobertAtNineMinds:
## Summary
Ticket **alga0002049**: appliance operators (reported by Marc, serverstart managed IT) had no clear answer to "will the appliance upgrade itself from 1.1.x to 1.2.x, or do I have to act?" — and unless they opened Manage → Updates, nothing in the product told them an update existed.

## UI (`ee/appliance/status-ui`)
- The status page now polls `/api/manage/status` on its own slow cadence (60s) and shows an **"Update available" banner** across the status tabs with an "Open Manage" CTA — covering both app updates and control-plane upgrades, with the available version and channel.
- The sidebar **Manage** button and the Manage **Updates** / **Control-plane** sub-tabs get an indicator dot when something is available.
- The Updates tab now states explicitly that the appliance checks the channel but **never applies updates automatically**.

## Host service (`ee/appliance/host-service`)
- `server.mjs`: cache the release-manifest resolve (60s TTL, keep last-known-good across transient registry failures) so the added status-page polling doesn't hit the OCI registry on every request — mirroring the existing `resolveControlPlaneRef` cache.

## Docs (`ee/docs/appliance`)
- Operator's manual: new **"Updates And The Upgrade Path"** section spelling out auto vs manual per layer — app updates are operator-initiated (1.1.x → 1.2.x is the same one-click channel update once published); the control plane refreshes at boot when the registry is reachable (holds last-good, never rolls back — per #2817) and can be upgraded in place; OS/k3s remain manual in v1.
- Quick-start and docs README updated to match.

## Testing
- `next build` of the status UI passes (TypeScript clean).
- Host-service tests: 79/80 pass; the one failure is the Docker-dependent fresh-install smoke test, which cannot run in this environment (no Docker daemon) and is unrelated.

https://claude.ai/code/session_01RtcuGv6AE4HFBH8nPKdvwX

---
_Generated by [Claude Code](https://claude.ai/code/session_01RtcuGv6AE4HFBH8nPKdvwX)_

### DIFF
--- a/ee/appliance/host-service/server.mjs
+++ b/ee/appliance/host-service/server.mjs
@@ -498,6 +498,31 @@ const manageKube = {
   }
 };
 
+// Resolve the channel's current release manifest (best-effort, cached 60s).
+// /api/manage/status is polled by the Manage view and, since the Overview
+// gained an "update available" banner, by the status page too — without this
+// cache every poll would hit the OCI registry twice. On a resolve failure the
+// last-known-good value is kept so update availability doesn't flap during
+// transient registry hiccups.
+// LEVERAGE: pattern appliance-channel-digest-drift — same TTL/last-good shape
+// as resolveControlPlaneRef below.
+const RELEASE_RESOLVE_TTL_MS = 60_000;
+const releaseResolveCache = new Map();
+function resolveReleaseManifestCached(reference, options = {}) {
+  const key = `${options.registryHost || ''}|${options.releaseRepository || ''}|${String(reference || 'stable')}`;
+  const cached = releaseResolveCache.get(key);
+  if (cached && (Date.now() - cached.at) < RELEASE_RESOLVE_TTL_MS) {
+    return Promise.resolve(cached.value);
+  }
+  return resolveReleaseManifest(reference, options).then((value) => {
+    releaseResolveCache.set(key, { value, at: Date.now() });
+    return value;
+  }).catch((error) => {
+    if (cached) return cached.value;
+    throw error;
+  });
+}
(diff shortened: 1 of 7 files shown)


RobertAtNineMinds:
Re: the AI review's "Should Fix" items — disposition:

1. **Interactive element IDs** — the Manage sub-tab buttons already carry `id={'manage-tab-' + value}` (pre-existing, unchanged by this PR; the diff shows them as moved context). The one new interactive element without an id is the banner's "Open Manage" button; the appliance status-ui doesn't follow the main app's UI-reflection convention anywhere (recovery, toolbar, and refresh buttons are all id-less today), so leaving it consistent here and deferring a sweep to a follow-up if wanted.

2. **Stale "update available" on persistent failures** — deliberate, and mirrors the existing `resolveControlPlaneRef` last-good behavior. The banner is informational only; "Run update" re-resolves the channel live and fails loudly, so a stale banner can never apply a stale release.

3. **Unbounded cache Map** — the key space is bounded in practice to the configured channel(s) (`stable`/`nightly`) with the registry host/repo fixed by the persisted release selection — a handful of entries for the life of the process, same shape as the CP resolve cache beside it.

On the suggestions: the appliance status-ui is a standalone static Next surface that doesn't consume the main app's shared component library or i18n (all of its copy is literal English), and "pick it up at boot" matches the post-#2817 `bootstrap-control-plane.sh` behavior (boot applies the channel's current image when reachable, holds last-good otherwise).

---

PR #2859, opened 2026-07-07
fix(billing): recover orphaned superseded recurring service period

RobertAtNineMinds:
## What this fixes

A recurring billing obligation was **superseded but never replaced**, silently dropping ~40.8 approved hours out of invoicing.

For the Nine Minds tenant, **AI Med Consult**'s June 2026 arrears obligation (`recurring_service_periods`, `period:2026-06-01:2026-07-01`) was flipped `generated → superseded` on 2026-06-24 with **no `generated` successor and no invoice**. Because the invoice generator excludes `superseded`/`archived` rows (`shared/billingClients/... invoiceGeneration.ts`, `whereNotIn('lifecycle_state', ['archived','superseded'])`), that period can never be invoiced — the approved hours are stranded even though the contract line, service, and time entries are all valid.

This is not isolated: a platform scan found **9 orphaned obligations across 3 tenants** with the same signature (latest revision `superseded`, nothing billed, invoice window already open).

## This PR: surgical data recovery

An idempotent, **tenant-scoped** migration that mints a `generated` successor at `revision = max(revision)+1` for orphaned periods (superseded tail, no active/billed sibling, invoice window opened), with `supersedes_record_id` pointing at the orphan. This is exactly the row shape the engine produces when regenerating over a superseded ledger (see `regenerateRecurringServicePeriods.test.ts` "assigns new records a revision above superseded ledger history"), so materialization/invoicing treat it as a normal open period.

- Scoped to the Nine Minds tenant; no-op elsewhere and on re-run.
- Verified by dry-run against production: selects exactly the June obligation, produces a constraint-valid `...:r2` generated row.
- Recovers the June hours so they invoice in the July window.

## Root cause (mechanism) — follow-up, not in this PR

The orphan is created by `regenerateRecurringServicePeriods` supersede-without-successor behavior: existing rows are paired to candidates **positionally** (`candidateIndex`), and an in-coverage period with no candidate is superseded. When a repair/schedule-change run's candidate window starts after (or omits) the current open period, that period is superseded with nothing to replace it.

Yesterday's `7d41ef4dde` ("stabilize recurring period repair regeneration") added `candidateCoverageEnd` + revision-continuity, which preserves the **tail** beyond coverage — but an open period **inside** claimed coverage with no candidate is still superseded (by design, per the new tests). So the durable forward fix belongs on the **caller side** (the anchor / coverage computation that decides which candidates get generated), and should be coordinated with that change. Deliberately **not** force-pushing a `regenerateRecurringServicePeriods` rewrite here, since it would conflict with the just-landed coverage model.

Recommend a follow-up to (a) ensure the regeneration anchor always covers the earliest surviving unbilled period, and (b) add a post-condition that refuses to supersede an unbilled period without a same-period successor — turning silent under-billing into a loud failure.

### DIFF
--- a/server/migrations/20260707120000_recover_orphaned_recurring_service_period.cjs
+++ b/server/migrations/20260707120000_recover_orphaned_recurring_service_period.cjs
@@ -1,0 +1,112 @@
+'use strict';
+
+/**
+ * Surgical recovery for recurring_service_periods obligations that were left
+ * "superseded but not replaced" -- i.e. their latest revision is `superseded`,
+ * nothing was ever billed for that period, yet the invoice window has already
+ * opened. Because the invoice generator excludes `superseded`/`archived` rows,
+ * such a period silently drops out of billing (an orphan): the approved work in
+ * that service period can never be invoiced even though the obligation is valid.
+ *
+ * This happens when a schedule regeneration / repair run supersedes the current
+ * open period without materializing a `generated` successor for it (see the
+ * positional-merge / coverage-anchor behaviour in
+ * shared/billingClients/regenerateRecurringServicePeriods.ts). The forward fix
+ * for the *mechanism* is tracked separately; this migration only repairs data
+ * that is already stranded.
+ *
+ * Recovery mints a fresh `generated` successor at revision = max(revision)+1 for
+ * the affected (schedule_key, period_key), pointing `supersedes_record_id` at the
+ * orphaned row. This is exactly the row shape the engine itself produces when it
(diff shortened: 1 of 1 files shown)


RobertAtNineMinds:
Superseded: recovered the single orphaned period with a direct, verified insert (generated r2, supersedes 4fc6981d) rather than a migration. Forward-mechanism fix (caller anchor/coverage in regenerateRecurringServicePeriods) remains a separate follow-up.

PR #2888, opened 2026-07-08
Stock Units: adversarial design review (resolve UUID columns, empty state, status colors, filter)

RobertAtNineMinds:
Adversarial design review of the **Stock Units** screen (`/msp/inventory/units`), driven by a six-lens antagonist critic team (voice, pixels, IA, intent, gaps, low-effort). Every change was verified live against the running screen and committed on its own.

## The screen's job
An MSP tech comes here to (1) find a serialized unit by serial/MAC, (2) see its state — status, location, client, warranty — and (3) inspect its movement history.

## Fixes (correctness — unanimous or near-unanimous across critics)
1. **Location & Client columns showed raw FK UUIDs.** The two columns that answer "where is it / who has it" dumped `location_id`/`client_id`. A `locationName` resolver already existed but was only used in the history dialog; Client had none. Resolved both at the reader layer (shared `stockUnitsWithNames` LEFT JOIN → `location_name`/`client_name`, applied to list + both search readers); table renders names via the kit `ClientNameCell`.
2. **CSV export contradicted the table** — raw UUIDs, omitted the Client column, raw status enum. Now exports the same resolved/humanized columns the grid shows.
3. **No empty/zero-results state** — a failed search rendered a bare header row (looked broken). Added `EmptyState` (search-specific message + Clear action), matching the sibling `StockOverview`.
4. **Status badge colored only 4 of 8 statuses** — allocated/in_transit/on_loan/returned all fell to the same grey as retired. All 8 now have distinct semantic variants; the fallback is a neutral catch-all.

## Polish
5. Normalize MAC display (`88AACC001122` → `88:AA:CC:00:11:22`), monospace identifiers, `—` for empty cells.
6. House-consistency: unit-count subtitle under the H1, right-aligned + non-sortable Actions column, fixed a stray `text-xs` in the dialog.
7. Swap the hand-rolled `Input` + separate Clear button for the kit `SearchInput` (icon + inline clear).
8. Tighten microcopy: actionable error toasts, example-format search placeholders, drop a trailing period.

## Fill
9. **Status filter** — `listStockUnits` supported it but the UI didn't; added a Status select so "show me all in-RMA / on-loan units" is one click. Narrows the loaded set client-side; count, export, and empty-state follow it.

## Declined / out of scope (deliberate)
- **Keep** the Warranty column (correctly built; empty only in dev data — populates in prod), the MAC column (identity + search key; normalized instead), and Refresh (working control).
- **Proposals, not built:** collapse Location+Client into one "Where" column (structural redesign); surface `advisoryLookup` bulk lookup / `listStockMovements` ledger / extra dialog fields (need new UI or belong on another surface).

## Verification
Type-clean (`types` + `inventory`). Every iteration exercised live via the browser (name resolution, CSV blob contents, zero-results + Clear, all 8 badge colors via temporary dev-data statuses since reverted, MAC normalization, SearchInput search/Enter/clear, Status filter incl. zero-match empty state). Light automated coverage per house norms; the rest smoke-validated live.


RobertAtNineMinds:
---

## Follow-up: MSP-owner usability pass (3 backed fills)

A teammate role-playing an MSP owner worked the live screen and flagged that it read as a bare serial-number ledger. Three of her gaps were backed by data already in hand and are now added:

- **Product column** — every row was a bare serial with no way to tell a firewall from an SSD. Joined `service_catalog` in the shared reader; the grid (and CSV) now shows the product name.
- **Unit Cost column + total inventory value** — cost was buried per-unit in the History dialog. Added a right-aligned Unit Cost column and a header total ("N units · $X value") that follows the active filter, so an owner can see cash tied up (and per-client value when filtered).
- **Client filter** — "what does this client have?" was unanswerable (search was serial/MAC only). Added a Client filter, built from the clients present in the loaded units, composing with the Status filter.

Deliberately **not** done here (surfaced as proposals): making the screen editable, richer RMA detail (vendor/RMA#/dates), and the larger question of **Stock Units vs. Assets** — the delivered stock unit becomes a much richer Asset record, so the warehouse/deployed handoff and duplicate hardware identity are a product/IA decision, not a design-pass fix.

PR #2889, opened 2026-07-09
fix(billing): stop per_unit inventory rows from crashing the service read path

RobertAtNineMinds:
## Problem

Two screens crashed on any tenant that has inventory products:

- **Sales Order** line-item picker → "No service found" for every search
- **Billing → Invoicing → Manual Invoice** → 500, "Failed to load data" / "No clients found"

Both depend on `getServices`, which validated every `service_catalog` row through `serviceSchema` with a throwing `.map(schema.parse(...))`. Inventory legitimately stores products as `service_catalog` rows with `billing_method = 'per_unit'` (and `unit_of_measure = null`) — see `packages/inventory/src/lib/materials.ts`. The T013 hard cutover deliberately purged `per_unit` from billing's canonical vocabulary (`fixed`/`hourly`/`usage`), so a **single** per_unit row threw a `ZodError` and aborted the entire list load. On Manual Invoice the failure also discarded the client list, since services and clients load in one `Promise.all`.

Console at repro:
```
Failed to load services: ZodError … "received":"per_unit" … invalid_enum_value
```

## Fix — read-path resilience, not a vocabulary change

`parseServiceReadRows` validates each row with `safeParse` and drops rows that billing's `IService` contract can't represent, so the canonical services still load instead of the whole call throwing. All three read sites (`Service.getAll`, `Service.getByCategoryId`, paginated `getServices`) route through it. Skipped rows are **logged** (count + `service_id`s), not silently swallowed — so a per_unit product not appearing in a billing-sourced picker is discoverable rather than a mystery.

### Deliberately NOT touched (hard-cutover intent preserved)
- `baseServiceSchema.billing_method` stays `z.enum(['fixed','hourly','usage'])`
- `IService`, `financialSchemas.ts`, `contractLineSchemas.ts` unchanged
- `billingInterfacesCutover.static.test.ts` unchanged and still passing

## Known follow-up (out of scope)

The sales-orders page calls `getServices(1, 999, { item_kind: 'any' })` — it *wants* products in the picker to sell them. With this fix, genuinely `per_unit` inventory products are excluded (they can't be represented as a billing `IService`). Surfacing sellable inventory products belongs in an **inventory-owned product reader**, not a widening of billing's per_unit-free contract. This is the same Stock-Units-vs-Assets / product-duplication IA question flagged in the inventory review — tracked separately.

## Verification

- **Live (devstack, port 3477):** invoicing tab **500 → 200**; sales-order service picker returns the **12 canonical services** (was 0); log-and-skip fires for the **4 `per_unit` products** in the seed.
- `packages/billing` typecheck: pass
- `packages/billing/tests/serviceReadRows.validation.test.ts` (new regression: one per_unit product row skipped, valid service still returned): pass
- `server` `billingInterfacesCutover.static.test.ts` (cutover guard, unmodified): pass


RobertAtNineMinds:
Rebased onto `main` (the branch had diverged with real conflicts in 9 files; all resolved) and addressed the review findings.

**Blocking findings**

1. *`parseServiceReadRows` could silently drop any unparseable row* — **fixed.** It now skips a row **only** when `billing_method === 'per_unit'` (the single legacy value billing's `IService` cannot represent — migration `20260323120000_normalize_per_unit_to_usage` rewrote existing rows but added no CHECK on `service_catalog`, so stragglers are possible). Every other validation failure calls `serviceSchema.parse(...)` and throws, preserving fail-fast. Added tests asserting a malformed `billing_method` and a missing required field both throw rather than vanish.

2. *Tenant isolation in `listInvoiceableSalesOrdersForBilling`* — **not an issue.** The query filters `.where('so.tenant', tenant)` and both joins carry an `andOn('*.tenant', '=', 'so.tenant')` predicate; there is no cross-tenant path.

3. *Permission gap on invoice generation* — **not an issue.** `generateInvoiceForSalesOrder` checks `hasPermission(user, 'sales_order', 'update')`; the composed fulfill/drop-ship actions enforce it too. `sales_order` is a seeded RBAC resource (`server/seeds/dev/47_permissions.cjs`).

**Merge reconciliation with main**

- #2892 (actions return `permissionError`/`actionError` instead of throwing): the kit actions this PR added were converted to the same contract, so `kitActions` is uniform. `KitManager` checks the returned error shape at every call site (a `catch` would never fire post-serialization), and its local `mapKitError` was replaced by the shared `kitActionErrorFrom` mapper, extended with the new sentinels and made to distinguish a SKU unique-violation from a duplicate component.
- "fix inventory currency handling": hardcoded `'USD'` fallbacks replaced by the tenant `defaultCurrencyCode`, and the sales-order line price now uses `CurrencyInput`, with this PR's kit price-override semantics preserved.

**Verification**

- `packages/billing` and `packages/inventory` typecheck clean; `server` typechecks clean (0 errors).
- Billing unit suite: 789 passed / 9 failed, where the 9 failures are identical on clean `main` (pre-existing, unrelated). New `parseServiceReadRows` tests pass.
- Inventory unit suite: the only non-passing files are DB-integration tests that need `server/.env.local` — same set fails on clean `main`. All kit-pricing/component unit tests pass.

PR #2983, opened 2026-07-19
fix(migrations): close NULL is_active hole in single-default client location dedupe

RobertAtNineMinds:
## Problem

The `20260718234058_enforce_single_default_client_location` migration failed in production:

```
could not create unique index "ux_client_locations_default_per_client"
Key (tenant, client_id)=(55f6a1b8-..., 3cda6833-...) is duplicated.
```

`client_locations.is_active` is a nullable boolean. The migration's cleanup steps only handled `is_active = false` (demote) and `is_active = true` (dedupe), so a duplicate default row with `is_active IS NULL` escaped both passes and broke the unique index build.

## Fix

The dedupe pass now ranks **all** remaining `is_default = true` rows regardless of `is_active`, keeping one per `(tenant, client_id)` — active rows preferred, then most recently updated, with `NULLS LAST` on timestamp tiebreakers so NULL-timestamp rows don't win ties.

Safe to edit in place: the migration never completed anywhere and runs in a single transaction, so it reruns from the top.

## Verification

Ran against a local Postgres with seeded scenarios: active default + NULL-active default (the production case), two NULL-active defaults with NULL timestamps, two active defaults, inactive default needing re-promotion, and a client with no default. Fixed SQL cleans all cases and the index builds; the original SQL reproduces the exact production error on the same data.

### DIFF
--- a/server/migrations/20260718234058_enforce_single_default_client_location.cjs
+++ b/server/migrations/20260718234058_enforce_single_default_client_location.cjs
@@ -1,5 +1,25 @@
 
 exports.up = async function up(knex) {
+  // On Citus, create_distributed_table() leaves the pre-distribution rows in the
+  // coordinator's local heap. They are invisible to all Citus-planned queries
+  // (including the dedupe UPDATEs below) but CREATE INDEX still scans them, so
+  // stale duplicates fail the unique index build. Truncate the leftover local
+  // data first; this cascades to local heaps of FK-referencing tables and never
+  // touches distributed (shard) data.
+  const citusInstalled = await knex.raw(
+    "SELECT to_regclass('pg_catalog.pg_dist_partition') IS NOT NULL AS has_citus"
+  );
+  if (citusInstalled.rows[0].has_citus) {
+    const distributed = await knex.raw(
+      "SELECT 1 FROM pg_dist_partition WHERE logicalrelid = 'public.client_locations'::regclass"
+    );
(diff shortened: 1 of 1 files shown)


RobertAtNineMinds:
Second failure (with the NULL-is_active fix deployed, migrate sha 72010d2381) had a different root cause: **stale Citus local heap data**.

Production `client_locations` is Citus-distributed (single-node, all 32 shards on the coordinator). `create_distributed_table()` left the pre-distribution rows in the coordinator's local heap — invisible to every Citus-planned query (the live data has exactly one default for the failing `(tenant, client_id)` pair, and zero duplicates anywhere), but `CREATE INDEX` still builds over the local parent heap, where an old duplicate default from before distribution (Aug 2025) survives. Telltale sign: the error's index name has no `_<shardid>` suffix, so it came from the coordinator-local build, not a shard.

Reproduced exactly on the citus-test sandbox (Citus even hints: *"The local data in the table is no longer visible, but is still on disk… run truncate_local_data_after_distributing_table"*), and verified the full migration sequence passes in one transaction after adding that call. It cascades to FK-referencing tables' local heaps only (tickets/contract_lines/assets also carry stale local data) and never touches shard data. No-op on plain Postgres/CE.

Note for the future: any migration adding a unique index/constraint to a distributed table can hit this same landmine on other tables with leftover local data.

PR #3037, opened 2026-07-28
feat(credits): derive credit balance from the ledger; delete reconciliation

RobertAtNineMinds:
Stacked on #3035 (which stacks on #3034).

## What

`clients.credit_balance` was a cache of a subtraction, kept in sync by ~15 writers, and the entire credit-reconciliation feature existed to detect and repair its drift. This PR makes the balance **derived** — SUM of non-expired `credit_tracking.remaining_amount`, defined once in `packages/billing/src/lib/creditBalance.ts` — and deletes the reconciliation feature wholesale, since a value that is computed from the ledger cannot disagree with it.

### Correctness by construction

- **Prepayment issuance moves to finalization.** Previously the transaction + tracking rows were written at invoice *creation* and the balance at *finalization* (and manual `isPrepayment` invoices bumped the balance with no tracking row at all — the exact `missing_credit_tracking_entry` class reconciliation hunted). Now finalization writes the `credit_issuance` transaction + `credit_tracking` row atomically; a draft prepayment grants nothing. A new nullable `invoices.credit_expiration_date` carries a manually chosen expiration from create to finalize; otherwise client/default billing settings are resolved at issue time.
- **All writers deleted**: apply, manual expire, transfer, void, invoice deletion, project-deposit rollback, expiration job, API createTransaction/transfer/bulk ops now touch only ledger + tracking rows.
- **Readers derive**: client API responses keep `credit_balance` as a computed field (correlated subquery — external consumers see no schema change), client stats/filters, financial analytics, account-balance report, and the billing-overview report metric all sum `credit_tracking`. Application gates are now currency-scoped, fixing the old cache's silent cross-currency addition.

### Deleted surfaces

Reconciliation actions + fix actions, reporting report actions/models, the credits-page Reconciliation tab (page is now a single table — `CreditsTabs` removed), `/api/v1/financial/reconciliation*` + `credits/validate` endpoints (controller, service methods, zod schemas, OpenAPI, route inventory), pg-boss + Temporal maintenance schedules, `credit_reconciliation_reports` table, locale sections in all 10 languages, and the reconciliation test suites.

### Migration

`20260728120000` adds `invoices.credit_expiration_date`, drops `clients.credit_balance` (down rebuilds it from tracking), drops `credit_reconciliation_reports`. Plain DDL, dual-target CE/Citus safe (no unique indexes involved).

## Testing

- `tsc --noEmit` clean: packages/billing, reporting, jobs, clients, types, db, integrations, onboarding, and server.
- Server unit suites for billing/jobs/api green after updating tests pinned to cache behavior (finalization kind classification, prepayment period policy, expiration handler, 6 FinancialService tenant-scoping contract tests, credits i18n suites, lang-pack validation — which also caught missing `addCredit.description` translations, now added in all languages).
- Failures left behind are environmental only: suites needing `rg` (absent here) and worktree module-resolution (`next/server`) failures that pass in the main checkout; unrelated domains (email, teams, mobile auth) fail identically without these changes.
- Not yet exercised against a running stack — recommend a smoke pass on grant → apply → transfer → expire → void before merging up the stack.

Also fixes a pre-existing missing `applyCreditToInvoiceInternal` import in server `InvoiceService`.


RobertAtNineMinds:
Closing rather than merging: this PR's content already landed on main via the rebased #3034 → #3035 → #3037 chain (main's fc062e4b36 is the same change under a different SHA). `packages/billing/src/lib/creditBalance.ts`, migration `20260728120000_derive_credit_balance_drop_cache_and_reconciliation.cjs`, the reconciliation deletion, and both `direct_grant` CREDIT_NOTE_CREATED schema fixes are all present on main today.

After merging origin/main, the net diff against main was only 30 stray `.pi-subagents/artifacts/*` files (gitignored at .gitignore:236, committed accidentally in 9aa86bbf3a) plus version drift — so merging would contribute 3,515 lines of agent transcripts and nothing else. CI was taken fully green (23 pass) to confirm there were no real defects hiding behind the staleness; all four original failures were branch-staleness artifacts.

PR #3110, opened 2026-08-05
fix: count hardware in every one-time total

NatalliaBukhtsik:
The dashboard's Opportunity snapshot summed nrr_cents alone while the
Pipeline table showed nrr + hardware, so the same deal read $600 in one
place and $1,000 in the other. One-time value now has a single
definition (oneTimeCents) that the snapshot, the pipeline table, the EE
forecast band and the seller rollups all route through, and the
snapshot spells out the hardware share when there is one.

The Hardware line was hiding behind the looking-glass all along; now it
comes to tea with the rest of the numbers. 🐇🔧💵


NatalliaBukhtsik:
Closing — this was opened from a stray branch by a mobile session. The two commits (9e61515eff, 063360a6f6) have been fast-forwarded onto `fix/fix_opportunity_reporting_and_streamline` and are now part of #3107.

PR #3120, opened 2026-08-06
alga0001984: POST /api/v1/invoices and /contracts return 500 — missing audit columns + API-key identity never established

RobertAtNineMinds:
## Problem

API-key-authenticated `POST /api/v1/invoices` and `POST /api/v1/contracts` returned generic 500 `INTERNAL_ERROR` responses even after passing Zod validation. Two root causes:

1. **Missing audit columns (Postgres 42703).** `ContractLineService.createContract` stamped `created_by`/`updated_by` via `addCreateAuditFields` and inserted into `contracts`, which has no such columns. `InvoiceService.create` did the same against `invoices`. It also targeted a phantom `invoice_line_items` relation — the canonical line table is `invoice_charges` (exposed to the product as the `invoice_items` view).
2. **API-key identity never established.** `ApiBaseController.authenticate` looked up the API-key user but never ran the request under `runWithApiKeyUser`, so downstream `withAuth`-wrapped server actions (e.g. `generateInvoiceNumber`) resolved no user and threw.

## Fix (in the shared mechanisms, not per-endpoint patches)

- **`ApiBaseController`**: all authenticated v1 verbs (list/read/create/update/delete) now run inside `runWithApiKeyContext`, which wraps the handler in `runWithApiKeyUser` alongside the existing `runWithTenant`, so the API-key identity is visible to every downstream `withAuth` action. The same wrapper is applied in `withApiKeyAuth` (apiAuthMiddleware.ts), which serves the non-controller v1 routes.
- **`BaseService.filterAuditFields`** (packages/db): now target-table-aware — takes an explicit `targetTable` so hand-rolled inserts into a table other than the service's default filter audit fields against the *actual* target's columns instead of failing the insert. Used by the `contracts` and `invoices` insert paths; a one-time warning logs each skipped column.
- **`createContractSchema`**: made strict (unknown fields → 400 `unrecognized_keys` instead of silent discard); accepts `client_id`, `start_date`, `end_date` and resolves the owning client from `client_id` or the legacy `owner_client_id` alias (backward compatible). Missing both → 400 naming `['client_id']`. Contract rows are created with explicit `is_template=false` so they stay visible to non-template listings.
- **Migration `20260805120000`**: adds `contracts.start_date` / `contracts.end_date`; the date columns are `timestamptz` (`useTz: true`) so caller-supplied instants round-trip without UTC-offset drift.
- **Invoice items**: `invoice_id` and `total_price` are no longer required from the caller on create — the server assigns the id and computes `total_price` from `unit_price * quantity`, persisting to the canonical `invoice_charges` table.
- **Error mapping**: Postgres 42703 (undefined column) now maps to a sanitized `INTERNAL_CONFIGURATION` error; the underlying column/relation detail is logged server-side only.

### Review-gap follow-up (d5df0a46a5)

- **`listContracts` projection**: the explicit select list now includes `start_date`, `end_date` and `is_template`, and maps `owner_client_id -> client_id`, so `GET /api/v1/contracts` returns exactly the fields `POST /api/v1/contracts` writes (previously the list silently dropped the new fields and `client_id` entirely).
- **Invoice line-item amounts**: `net_amount` is now optional on create and derived server-side as `unit_price * quantity` when omitted; nullish `quantity`/`unit_price` fall back safely; zero-quantity items persist as 0/0/0.
- **Behavioral integration tests** added for contract create (incl. owner-client alias + strict schema), contract list projection, and invoice line-item derivation.

## Test coverage

- **8 integration tests** across `contractLineService.createContract.integration.test.ts`, `contractLineService.listContracts.integration.test.ts`, `invoiceService.createLineItems.integration.test.ts` — contract create/list projection and invoice line-item derivation against the bootstrapped test DB.
- **11 schema/unit tests** across `BaseService.auditFields`, `apiKeyUserContext`, `apiMiddleware.productAccess` (incl. the 42703 → INTERNAL_CONFIGURATION mapping), `contractCreateOwnerClientSchema`, `contractSchema.systemManagedDefault`, `invoiceCreateItemSchema`.

## Smoke test summary (recheck @ d5df0a46a5)

Recheck smoke of `fix/alga0001984-api-v1-post-500` @ d5df0a46a5 against the real running stack (compose `alga-psa-local-test`, app on :3366 from this worktree, restarted this run with `PORT=3366 npm run dev`). Environment: verified genuinely up (containers healthy, server answering, process cwd = worktree). Full-stack fidelity — level (a) throughout; no simulators/mocks. Logged into the real UI as glinda (fresh boot-banner password) via an agent-hosted browser pane created by sending `browser.createPane` over the agent wire protocol (the agent's restart at 14:28 had wiped its hosted browsers, and CLI browser commands only drive agent-hosted panes; this is the same mechanism the prior run used).

### FLOWS, ALL PASS

1. `POST /api/v1/contracts` x-api-key -> **201** (was 500), `client_id` echoed, `is_template=false`, `start_date`/`end_date` timestamptz exact round-trip; row confirmed in Postgres.
2. `POST /api/v1/invoices` x-api-key -> **201** (was 500); items omitting `net_amount`/`total_price` derived server-side (7000 from 1x7000), zero-quantity item persisted as 0/0/0, `created_by=glinda` on `invoice_charges` proving the API-key identity fix.
3. `GET /api/v1/contracts` projects `client_id`/`is_template`/`start_date`/`end_date` (the d5df0a46a5 list-projection fix).
4. 42703 -> 500 `INTERNAL_CONFIGURATION` live (was `INTERNAL_ERROR`), sanitized message + server-side diagnostic log with offending SQL.
5. Strict `createContractSchema`: unknown field -> 400 `unrecognized_keys` (no silent strip); missing `client_id` -> 400 naming the field.
6. Regression: `GET` clients/tickets/users/invoices all 200 under the new `runWithApiKeyUser` wrapper.
7. Tests: 8 integration (contract create/list projection, invoice line-item derivation) + 11 schema/unit incl. 42703 mapping, all green against the bootstrapped test DB.
8. UI: dashboard, billing client-contracts + invoices tabs render, and the Create Contract wizard opens — proving the UI create path is unregressed by the auth-wrapper change.

### SUSPICIOUS/NOTED

- **(a)** The billing UI was 500ing on `service_catalog.barcode` — a PRE-EXISTING schema-drift 42703 (migration `20260716120000` never applied because knex migrate is blocked by the corrupt ledger; the file is not part of this branch). Applied the ALTER + partial unique index manually to unblock UI rendering; the 42703 mapper is what now reports it as `INTERNAL_CONFIGURATION` instead of hiding it.
- **(b)** v1 `createContract` writes only the `contracts` header and creates NO `client_contracts` assignment row, so the assignment-driven UI `getContractsWithClients` list does not show v1-created contracts — expected per the contract-assignment model, not a regression.
- **(c)** One pre-existing smoke key `'alga0001984 live verify'` (from the earlier 000825 run) remains in the DB, left untouched.

### CLEANUP

All rows/keys created this run deleted and verified 0 smoke rows remain; worktree clean; evidence not committed. Fidelity notes: browser UI used for login/rendering/wizard evidence; the headline create assertions are API+DB level (which is where the defect lived).

Implementation plan: `docs/plans/2026-08-05-alga0001984-api-v1-create-plan.md` (committed at ae3b99a5af).

Evidence directory: `/tmp/alga-smoke-evidence/alga0001984-api-v1-post-500-20260806-recheck`


RobertAtNineMinds:
Reopening to trigger CI on current head (0010738e) — the two required checks (ext-v2 guard, circular deps) never ran for the newest commits.

PR #3164, opened 2026-08-14
docs: reflect invoice/sales-order PDF filing (2026-08-14)

NatalliaBukhtsik:
## Source PRs (alga-psa)

- **#3152** – [File generated invoices and sales orders as documents](https://github.com/Nine-Minds/alga-psa/pull/3152)

### `docs/billing/invoice_finalization.md`
- **Finalization Triggers section**: Clarifies that the "email sending" trigger is still not implemented for `finalized_at` purposes, but email now does publish the filed PDF to the client portal (the two concerns are now distinct).
- **Database Schema section**: Documents the new `document_associations` entity types (`invoice`, `sales_order`) and the three new provenance columns on `documents` (`source_template_id`, `source_template_version`, `rendered_locale`).
- **PDF Generation Integration section**: Adds a description of the refresh-until-issued document-filing model: each download re-renders in place while the invoice is MSP-only; `publishGeneratedDocumentsToClient` is called on email send to freeze the stored copy and make it visible in the client portal. Notes the legacy code snippet reflects the previous simplified implementation shape.
- **Implementation Status — ✓ list**: Adds invoice PDF filing, client-portal PDF consistency via `publishGeneratedDocumentsToClient`, and `invoice`/`sales_order` in `document_associations`.
- **Implementation Status — ⚠️ list**: Moves "email integration" here with a precise note: sending email publishes the filed PDF to the client portal but does not set `finalized_at`.
- **Implementation Status — ❌ list**: Renames the email-finalization item to "Automatic finalization on email send" to distinguish it from document publishing (which is now done).
- **Future Enhancements — Email Integration**: Marks PDF-filing-on-send as done (✓) and clarifies the remaining item is auto-setting `finalized_at` on email send.
- **Integration Tests section**: Adds two new test items covering the refresh-until-issued behavior and the `publishGeneratedDocumentsToClient` freeze.

## Needs Human Review

- **PR #3157 (fix(email): clarify outbound sender identity UX)**: Changes moved the From address field into the SMTP card and hid the "All other email identity" card for SMTP providers. The `docs/email/email_settings_access.md` SMTP field summary omits "From address" — please verify whether this list should be updated, or if the summary level is intentionally brief.

- **PR #3161 (fix(schedule): keep entry end after start)**: New UX where End shifts to preserve duration when Start is moved past it. No scheduling docs are in-scope paths, but a human should confirm.

- **PR #3159 (Fix/nightly run august13)**: No PR body; CI/nightly fix assumed. Please verify no behavioral change.

---
_Generated by [Claude Code](https://claude.ai/code/session_011v3ptJ9LMRqRnJkmtsWT6Q)_

### DIFF
--- a/docs/billing/invoice_finalization.md
+++ b/docs/billing/invoice_finalization.md
@@ -13,7 +13,7 @@ The invoice finalization system provides a way to mark invoices as finalized onc
 
 An invoice becomes finalized when:
 1. **Manual finalization** - User explicitly finalizes via UI action menu
-2. **Email sending** - Future feature (not yet implemented)
+2. **Email sending** - Future feature (not yet implemented — sending email now publishes the filed PDF to the client portal but does not set `finalized_at` or change invoice status)
 
 **Important:** PDF download does NOT automatically finalize invoices. Users must explicitly finalize invoices through the UI action menu.
 
@@ -50,6 +50,13 @@ The `invoices` table includes:
 ```sql
 finalized_at TIMESTAMP WITH TIME ZONE
 ```
+
+The `document_associations` entity-type check constraint now includes `invoice` and `sales_order`, allowing filed PDFs to be linked to their source entity and client.
+
+The `documents` table carries three new provenance columns (nullable, forward-only):
+- `source_template_id` — ID of the document template used to render the PDF
+- `source_template_version` — version of that template at render time
+- `rendered_locale` — locale resolved during rendering
 
 See [IInvoice interface](../server/src/interfaces/invoice.interfaces.ts) line 15 for TypeScript type definition.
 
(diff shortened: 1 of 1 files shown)


NatalliaBukhtsik:
Thanks for the review. Addressing each point:

**Should Fix 1 — filing failure swallowed:** This is explicitly intentional per the source PR (#3152) body: *"A filing failure is logged and swallowed rather than surfaced — the stored PDF itself is still usable for download/email even if filing fails."* The doc reflects this design accurately. Whether that failure should be surfaced to an audit trail instead is a product question for a human reviewer, not something I can change in documentation.

**Should Fix 2 — email send vs. `finalized_at`:** Also accurate as written. PR #3152 implements document publishing via `publishGeneratedDocumentsToClient` (makes the filed PDF visible in the client portal) but does not touch `finalized_at` or invoice status. The doc calls this "partially implemented" precisely to avoid that confusion — the email-integration item in Future Enhancements now clearly distinguishes ✓ document publishing from the still-pending auto-finalization. No change needed.

**Suggestion 1 — forward-only note for `document_associations`:** The new provenance columns on `documents` are already marked "(nullable, forward-only)" in the doc. The entity-type expansion on `document_associations` is a check-constraint migration, not a nullable column, so "forward-only" doesn't apply the same way — calling it that would be misleading.

**Suggestion 2 — separate subsection for invoice vs. sales-order:** The filing model is identical for both (same `PDFGenerationService.generateAndStore` path, same refresh-until-issued logic); separating them would duplicate prose rather than clarify. Leaving as-is.

No file changes from this review. Flagging for the human reviewer.

---

PR #3186, opened 2026-08-17
fix(migrations): unblock the hosted migration chain (Citus FKs + duplicate Bucket configs)

robertisaacs:
Root-fixes every blocker on the hosted combined-migration chain inherited from PR #3176, in dependency order. The chain first died on the single-node Citus `hour_blocks_invoice_fkey` FK; each fix uncovered the next blocker behind it, so this branch carries the whole sequence.

> The locale hour-block keys, tenant-deletion order, and jobs/unit-test repairs from the card title landed separately on `main` via #3179 (merge `de8dc2ac`). This branch's remaining diff is migration-only — the six commits below.

## Blocker 1 — Citus FK / colocation (`hour_blocks` chain)

**`20260813120000_create_hour_block_tables.cjs`**

1. Distribution used bare `create_distributed_table` with no colocation, so the composite FKs to `invoices` / `clients` / `service_catalog` landed in a different colocation group and were refused. All four hour_block tables now distribute `colocate_with => 'invoices'` (house pattern: `20260722120000_create_invoice_payments_table.cjs`).
2. `hour_blocks_invoice_fkey` declared bare composite `ON DELETE SET NULL` — which would null the `NOT NULL` `tenant` column on plain Postgres, and which Citus rejects outright when the distribution key is in the FK. Recreated per `20260719120000_fix_suppression_contact_fk_set_null.cjs`: column-targeted `SET NULL (source_invoice_id)` on plain PG 15+, plain `NO ACTION` on Citus. Safe because `hardDeleteInvoice` detaches `source_invoice_id` in-app first and tenant deletion removes `hour_blocks` before `invoices`.
3. `hour_block_allocations_entry_fkey` references `time_entries`, a local (non-distributed) table on Citus; the FK is now added only where enforceable — plain Postgres, or a Citus cluster with `time_entries` distributed.

**`20260815120000_link_hour_blocks_to_invoice_lines.cjs`**

4. `hour_blocks_source_charge_fkey` used column-targeted `SET NULL` unconditionally, which Citus also refuses. Where `SET NULL (column)` is unavailable the FK is skipped rather than downgraded to `NO ACTION`; `resolvePurchaseLineForBlock` already treats a dangling id as detached.

**New `20260817120000_fix_hour_blocks_invoice_fk_set_null.cjs`**

5. Drops and re-adds `hour_blocks_invoice_fkey` in the safe form for databases that already ran the original bare-`SET NULL` migration. No-op on fresh chains.

## Blocker 2 — duplicate legacy Bucket configs (`20260814090000`)

6. The member backfill died on a UNIQUE `(tenant, contract_line_id, service_id)` collision: production carries 11 groups (two tenants) of legacy Bucket configs that are application double-writes, not real buckets. `CANONICAL_BUCKET_CONFIGS` collapses each duplicate group to its earliest row before the pool backfill.
7. Pool-term backfill (step 3b) used `ON CONFLICT DO NOTHING`, which — on the `transaction: false` retry path the divergence guard itself creates — would discard a reconciliation and keep stale terms. Now merges the four term columns from the legacy config (the source of truth until this migration completes); pool identity and new-model columns are preserved.
8. Unmappable usage is quarantined into `bucket_usage_unmappable_archive` instead of deleted — the delete predicate could match real invoiced history on customer-managed appliances (config lineage was snapshotted 2025-10-08). `bucket_id` still goes `NOT NULL`, but nothing is destroyed and `down` leaves the archive in place. The usage-to-assignment match is also now anchor-ranked rather than a raw overlap, so a healthy mid-period contract switch no longer matches two pools and aborts.

## Blocker 3 — `run_command_on_shards` format() overflow (`20260814090000` step 3i)

9. `run_command_on_shards` substitutes a single shard-qualified name via `format(command, shard_name)`, so a two-`%s` body overflowed it (`too few arguments for format()`). Split into one `ALTER` per call (house pattern: `20251205120000_...`); `down` fixed identically. Reached only on the hosted shape where `bucket_usage` is distributed — which is why three deploys got past local verification, where it stays a plain table.

## Blocker 4 — credit restriction-mode constraints (`20260816120000`)

10. The `SET NOT NULL` / `SET DEFAULT 'all'` on `credit_service_type_restriction_mode` failed on the hosted table's shard state despite a `WHERE`-less backfill that always yields a value (root cause not diagnosed — row-count divergence ruled out). Neither constraint is load-bearing: `resolveCreditPolicy()` normalizes NULL/absent/unknown to `'all'`, exactly what the backfill writes. New `20260817130000_drop_credit_restriction_mode_constraints.cjs` drops both CHECK/NOT NULL constraints on `default_billing_settings` and `client_billing_settings`, leaving the columns nullable with an `'all'` default.

## Blocker 5 — CI guards on the new objects

11. `bucket_usage_unmappable_archive` (the quarantine table added in blocker 2) carries a `tenant` column, so `scripts/validate-tenant-management.ts` requires it in `TENANT_TABLES_DELETION_ORDER`. Added next to `bucket_usage` in `ee/temporal-workflows/src/activities/tenant-deletion-activities.ts`; it is a pure leaf (no FKs in or out, by design), so its deletion position is unconstrained.
12. The drop migration was renamed `20260818090000 → 20260817130000` so `custom-rules/migration-filename` (ESLint) no longer flags a future timestamp; still ordered after `20260816120000`, which it undoes.

## Smoke-test results — all PASS

Verified at HEAD `89fd8f03` against a fresh single-node **Citus 12.1** (`shard_count=4`):

- **01** — Full combined CE+EE chain: **994 migrations, one batch, exit 0.** 237 distributed + 11 reference tables; all 17 asserted tables distributed; **`hour_blocks` colocated with `invoices`** (the root defect); Citus-side `hour_blocks` FKs carry no `SET NULL`; restriction-mode CHECKs gone, columns nullable with `'all'` default.
- **02** — With `bucket_usage` distributed + colocated (the hosted shape CI never exercises), replaying `20260814090000`'s `up()` runs the `usageCitusDistributed` branch cleanly — one command per `run_command_on_shards` call; coordinator and shard types both `numeric(12,2)`.
- **03** — Live dev stack (port 5472): `migrate:status` = **No Pending Migration files Found** at HEAD; restriction-mode CHECKs absent; `hour_blocks` FKs keep column-targeted `SET NULL` on plain PG.
- **04** — App boots on port 3657: `200` → `/auth/msp/signin`, MSP Sign In renders.

Unit suite green earlier the same day (`UNIT_EXIT=0`); the five commits after that run are migration-only and are re-validated by the fresh chain in 01.

**Evidence directory:** `/tmp/alga-smoke-evidence/root-fix-3176-ci-blockers-20260817-205913/` (`SMOKE-SUMMARY.md`, `01-citus-full-chain.log`, `02-citus-distributed-bucket-usage-replay.log`, `03-livedev-5472-schema.txt`, `04-app-3657.png`).


robertisaacs:
## Second blocker: duplicate legacy Bucket configs (8f5d127)

The hour-block FK fixes got the hosted chain past `hour_blocks_invoice_fkey`; it then died in `20260814090000` on `contract_line_bucket_services_tenant_contract_l_58536c1e_118970`. The `ON CONFLICT` there names the primary key, but the violated index is the table's *other* unique — `UNIQUE (tenant, contract_line_id, service_id)`. Since `bucket_id` is seeded from `config_id`, a PK collision cannot happen, so two legacy Bucket configs must share a `(contract_line_id, service_id)`.

### Where the data came from

`20250318200000` declared `UNIQUE (plan_id, service_id, tenant)` on `plan_service_configuration`. `20251008000001:850` recreated the table as `contract_line_service_configuration` with only `PRIMARY KEY (tenant, config_id)` and backfilled it with a bare `ON CONFLICT DO NOTHING`. Nothing has enforced one-config-per-(line, service) since October 2025.

Production has 11 such groups across 2 tenants. Every one is a pair written 3–40ms apart with **identical** `total_minutes` / `overage_rate` / `allow_rollover` / `billing_period`, differing only in the second row having `custom_rate` and `quantity` NULL — an application double-write, not two intentional buckets. Collapsing is therefore lossless (`dup_groups = 11, dup_groups_with_divergent_terms = 0`).

The double-write itself and the missing unique constraint are a **separate follow-up** — this PR only makes the migration survive the data.

### A third blocker was queued behind it

Step 3d's orphan guard would have failed the next deploy: only 2 of 5 production `bucket_usage` rows mapped.

- One real customer row failed purely because its period opens 2026-06-28 while the client contract assignment starts 2026-07-01 — the anchor was a point test on `period_start`. It is now an interval overlap against `[period_start, period_end]`, which is what a usage *period* means.
- Two rows in a 2024-09 test tenant have `service_catalog_id` NULL on lines carrying no Bucket config. Nothing can hold them and `bucket_id` goes NOT NULL, so 3e deletes rows that have no Bucket config for their service **and** none for their line — never billable under the old schema either. Rows that *do* have a candidate config and fail the client/window join still hard-fail. Every deletion is logged with its minutes.

### Three Citus shapes had to change

All latent in CI, because an empty chain never reaches them:

1. Correlated subqueries cannot read a CTE or subquery (`correlated subqueries are not supported when the FROM clause contains a CTE or subquery`) — the sweep and 3e became top-level anti-joins, and the inserts became distribution-key-qualified point writes, the same resolution the usage remap already used.
2. `COUNT(DISTINCT ...)` in `HAVING` either mistypes the push-down (`COALESCE types numeric and character varying cannot be matched`) or returns groups that fail the predicate. It is now projected in a grouped CTE and filtered in an outer `WHERE` — which also repairs the **pre-existing** ambiguity guard in 3d.
3. Row comparison under `COUNT(DISTINCT ...)` does not survive push-down, so terms compare as a concatenated text key.

### Verification

- Full **996-migration** combined CE+EE chain green from scratch on `citusdata/citus:12.1` (shard_count 4, coordinator holding shards).
- Re-run on a second Citus database seeded with production's exact shape — duplicate pair with identical terms, a usage row opening before its assignment, an unmappable NULL-service row. Result: 1 pool / 1 member at multiplier 1.000 keyed on the **earliest** config, early row mapped, unmappable row deleted, 0 unmapped.
- Guard confirmed both ways: passes on equivalent duplicates, fires on divergent ones.
- Every query also run read-only against the hosted cluster: 33 canonical configs, 11 stale pools swept, 0 ambiguous, 3 mapped, 2 deleted.

### Reviewer attention

The FK weakening from the first commit still stands (`NO ACTION` for `hour_blocks_invoice_fkey` on Citus, `hour_blocks_source_charge_fkey` skipped) — integrity there rests on app-level detachment. New here: the 3e deletion is the only step that destroys data, and the 44→33 pool sweep assumes any pool keyed on a non-canonical config is debris from the earlier failed run, which holds because both tables are created by this migration.

robertisaacs:
## Review response: hardened for appliances (92eab8a, 5bb35d5)

An adversarial review of `8f5d127` found three real issues and refuted six hypotheses. All three are now fixed.

### F3 — pool terms were never reconciled across runs (`92eab8a`)

`3b` used `ON CONFLICT DO NOTHING`, so a pool keyed on a legacy `config_id` that already existed from an earlier failed run kept whatever the config held *at that moment*, permanently, while the migration reported success. Production already carries 33 such pools from the run that died on the member insert.

It matters most on the retry path the divergence guard itself creates: that guard tells the operator to reconcile the configs and re-run, and `DO NOTHING` would discard the reconciliation. Now merges the four term columns; `contract_line_id` and the new-model columns are left alone.

Current production exposure is **nil** — 33 canonical pools, 0 disagreeing with their configs — so this does not change the pending deploy's outcome.

### F1 — `3e` no longer deletes; it quarantines (`5bb35d5`)

My justification for deleting was wrong. "No config today, so it was never billable" assumes configs are immortal, but the lineage was **snapshotted on 2025-10-08**: `20251008000001` backfilled `contract_line_service_configuration`, then `20251008000003` dropped `plan_service_configuration CASCADE`. Any bucket config retired before that snapshot is absent today while its `bucket_usage` rows survive. An appliance that ran buckets through 2024 and retired the plan in 2025 has real invoiced history matching the delete predicate. `bucket_usage.contract_line_id` compounds it — the old write path populated it unreliably, so the line half of the test can miss too.

Rows now move to `bucket_usage_unmappable_archive`. `bucket_id` still goes NOT NULL so they still cannot stay in `bucket_usage`, but nothing is destroyed and recovery is from a table rather than console scrollback. `down` deliberately leaves the archive in place — which also defuses the `down; up` round trip, where new-model usage with no legacy config behind it would otherwise have been deleted on the re-`up`.

### F2 — a regression I introduced in `8f5d127` (`5bb35d5`)

Widening to a raw interval overlap made **sequential replacement** look ambiguous. An assignment ending 06-15 and its replacement starting 06-16 both overlap a 06-01..06-30 usage period, so a healthy mid-period contract switch matched two pools and aborted the migration — a case the original point test handled cleanly, and one the existing tests miss (they only cover replacement on a *period boundary*, where inclusive bounds don't collide).

Candidates are now ranked: an assignment whose window contains `period_start` is preferred, because that is the anchor the old write path resolved the draw with. Overlap-only candidates remain as the fallback that rescues a mid-period contract *start*. A row is ambiguous only when two pools tie at its best rank — genuinely concurrent assignments. The error text now says that, instead of sending operators hunting for duplicate assignments that don't exist. Ranking in JS also collapses the guard and the mapping into one candidate set, so they can't drift apart the way two hand-written SQL queries could.

**Both ranks demonstrated on `citusdata/citus:12.1`:**

| usage row | `raw_overlap_pools` | `anchored_pools` | outcome |
|---|---|---|---|
| mid-period replacement | 2 — *`8f5d127` would have aborted here* | 1 | maps to the **original** contract's pool, where it was written |
| mid-period contract start | 1 | 0 | rescued by the fallback rank |

Full 996-migration chain green with both fixtures present: 3 pools, 3 members, 0 unmapped, 1 row archived with its minutes intact.

### Refuted

`client_contracts.start_date` / `end_date` are `date` in the live schema, not `timestamptz` — the window predicate is date-to-date with no session-TimeZone promotion, so the claimed TZ-dependence does not exist. Also refuted: invoice lineage (that config FK died with the same `DROP ... CASCADE`, and no config rows are deleted here), the winner rule (`created_at` NOT NULL, tie-broken by `config_id`), re-runnability (converges from every crash point), the sweep premise (`templateClone.ts` pools write no legacy row, so they never join `legacy`), and silent minute loss via a NULL `term_key` (all four compared columns are NOT NULL).

### Known and accepted

- A member-less phantom pool survives if an operator deletes a duplicate *config* row between runs — cosmetic, and tightening it would risk deleting legitimate `templateClone` pools.
- A seconds-wide TOCTOU between the ambiguity guard and the remap during a rolling deploy, since the migration is `transaction: false`.

robertisaacs:
## Third blocker: `too few arguments for format()` (8ff85f1)

Step 3i. `run_command_on_shards` substitutes the shard-qualified name with `format(command, shard_name)` — a **single** argument — so a body carrying two `%s` overflows it:

```
ERROR:  too few arguments for format()
```

Split into one `ALTER` per call, matching the house pattern in `20251205120000_change_template_tasks_estimated_hours_to_minutes.cjs`, which already alters one column per call on the same table. `down` carried the identical two-`%s` body and is fixed the same way. A scan of every dollar-quoted `run_command_on_shards` body in the repo finds **no other instance** — these two were the only ones.

### Why local verification kept missing this

**No migration distributes `bucket_usage`.** It is registered tenant-scoped in `migrations/utils/tenantDb.cjs:47`, but production distributes it out of band. So a locally-migrated database leaves it a plain table, and every `if (usageCitusDistributed)` branch in this migration — 3h's `SET NOT NULL` and 3i's widening — is dead code locally. Three green local runs told me nothing about either.

The rehearsal now runs `create_distributed_table('bucket_usage', 'tenant', colocate_with => 'contract_lines')` before the bucket migration, which is what finally exercised those branches. Worth knowing for any future migration that touches this table.

### Verification

- Minimal distributed table: the two-`%s` form reproduces `too few arguments for format()`; the split form returns success for **every** shard.
- Full 996-migration chain green on `citusdata/citus:12.1` with `bucket_usage` distributed.
- After `up`: coordinator catalog **and** all shards report `numeric(12,2)`, and a `42.75 / 3.25` round trip persists — the widening is real, not just a catalog edit.
- After `down`: both report `bigint` again, pool tables dropped, archive survives.

### Production state right now

The failed run left a clean intermediate state:

```
pools 33 | members 33 | usage_rows 3 | unmapped 0 | archived 2
bucket_id NOT NULL ✓   minutes_used/overage_minutes still bigint (3i never ran)
```

A re-run is a no-op through 3h (sweep finds nothing, pools merge, members conflict-ignore, no NULL `bucket_id` rows to map or archive) and resumes at the widening.

**Note for reviewers:** the archive added in `5bb35d5` earned itself here. Those 2 rows hold 110 and 35 minutes, and they are intact in `bucket_usage_unmappable_archive`. The `DELETE` it replaced ran in this very deploy and would have destroyed them before anyone saw the failure.

robertisaacs:
## Fourth blocker: credit restriction-mode constraints (89fd8f0)

The bucket migration now passes. The chain moved on to `20260816120000`, which added `credit_service_type_restriction_mode`, backfilled it, then failed on:

```
ALTER TABLE default_billing_settings
  ALTER COLUMN credit_service_type_restriction_mode SET NOT NULL,
  ALTER COLUMN credit_service_type_restriction_mode SET DEFAULT 'all'
- column "credit_service_type_restriction_mode" ... contains null values
```

### Honest status on the cause: not diagnosed

The backfill's `CASE` has no `WHERE` and always yields a value. Production has 49 rows, **all** with `credit_eligible_service_type_ids` SQL NULL, so every one should become `'all'`. A faithful knex replay against a distributed table with the same shape, shard count and row count **passes**. Per-shard counts sum to exactly the 49 the coordinator reports, so row-count divergence is ruled out — but nothing else is.

**This change avoids the constraint rather than explaining it.** Flagging that plainly: something about the hosted table's shard state is still unaccounted for.

### Why dropping them costs nothing

Both constraints only restated invariants the application already maintains:

- **NOT NULL** — `resolveCreditPolicy()` (`shared/billingClients/billingSettings.ts:316`) reads the tenant mode via `normalizeServiceTypeRestrictionMode(value, 'all')`, so NULL, absent, or unrecognized all resolve to `'all'` — exactly what the backfill writes. The tenant-side `DEFAULT 'all'` is kept, so new rows stay populated regardless.
- **The CHECKs** — `updateClientBillingSettings()` / `updateDefaultBillingSettings()` (`billingSettings.ts:185-208`) treat mode as the source of truth, null the ids for `'all'`/inherit, and derive mode from ids for legacy callers. The code comment literally says it does this "so the DB CHECK constraints hold." The read path only consults ids when mode is `'restricted'`.

The column, the empty-array normalization, and the backfill all stay. Only the NOT NULL and the two CHECKs go.

This also removes a **fifth blocker that was queued behind it**: the surviving `SET DEFAULT` is now its own `ALTER`, because Citus rejects an `ALTER` carrying two utility subcommands with `cannot execute multiple utility events` — which the original two-subcommand statement would have hit as soon as the NULL problem was solved.

### Convergence for databases that already installed them

`knex` will not re-run the edited migration where it already succeeded, so new `20260818090000_drop_credit_restriction_mode_constraints.cjs` drops them there. Every statement is `IF EXISTS` or gated on catalog state; `DROP NOT NULL` is guarded on `attnotnull` since it is not `IF EXISTS`-able. `down` is deliberately a no-op.

### Verification

On `citusdata/citus:12.1`, against **distributed** `default_billing_settings` and `client_billing_settings` (32 shards, 49 + 20 rows, some seeded `'[]'`):

| scenario | result |
|---|---|
| fresh install | `notnull=false`, 0 checks, 0 NULL tenant modes (confirmed per-shard), 20 NULL client modes (inherit, by design) |
| DB carrying the old constraints | converges to the same state |
| re-run on converged DB | clean no-op |

Full **997-migration** combined chain green from scratch, ending with no NOT NULL, no CHECKs, and `DEFAULT 'all'` on the tenant column.

PR #3265, opened 2026-08-26
fix(appliance): preserve Temporal connection binding

RobertAtNineMinds:
## Summary
- preserve the Temporal `Connection` class as the receiver of its static `connect` method
- prevent Appliance Console license redemption and application from failing before workflow submission
- add a behavioral regression test that exercises the SDK-style `connect -> this.lazy` call chain

## Verification
- `node --test ee/appliance/host-service/tests/appliance-license-workflow.test.mjs ee/appliance/host-service/tests/manage-engine.test.mjs` (20/20 passed)
- `node --check server/scripts/appliance-license-workflow.mjs`
- `node --check ee/appliance/host-service/tests/appliance-license-workflow.test.mjs`
- `git diff --check`


RobertAtNineMinds:
Closing because GitHub never computed mergeability or registered pull-request Actions after opened, reopened, and synchronize events. Replacing with a fresh PR from the same verified commit.

PR #3321, opened 2026-09-03
ci: raise integration-tests job timeout to 120 minutes

RobertAtNineMinds:
## What
Raise the job-level `timeout-minutes` on the `integration-tests` job in `.github/workflows/integration-tests.yml` from 90 to 120.

## Why
The integration suite has been running close to the 90-minute cap; the extra headroom prevents legitimate runs from being cancelled by the job timeout.

## Testing
Workflow-only change, no application code touched. Step-level timeouts for the non-blocking infrastructure steps are unchanged.


RobertAtNineMinds:
Superseded by a clean PR containing only the workflow change; this one accidentally carried unrelated local commits and was in a conflicting state.

PR #3339, opened 2026-09-05
fix(payments): verify Stripe payments webhook against tenant endpoint secret

RobertAtNineMinds:
## Problem

Customers (NVTS reported it) received Stripe "failing webhook endpoint" notices for `https://algapsa.com/api/webhooks/stripe/payments`.

Two bugs stacked:

1. The route verified every event against the single platform webhook secret. When a tenant connects their own Stripe account, Alga creates a webhook endpoint in *their* account and stores that endpoint's signing secret as a tenant secret, but the route never read it. Tenant-account events could never verify.
2. In production the platform secret was only present as an env var, never in Vault, and the route has no env fallback (removed in 394a6864ef, later restored only for the license service). Every delivery since then returned 503. The Vault side has been fixed operationally; this PR is the code side.

## Fix

- New `paymentsSignature.ts`: peek `tenant_id` from the unverified payload purely to pick candidate secrets (tenant endpoint secret first, then platform `stripe_payment_webhook_secret` / `stripe_webhook_secret`), then verify against each. Nothing from the payload is trusted until a signature matches.
- Logs record the tenant hint, sources tried on failure, and the matching source on success.
- Both Stripe webhook `route.ts` files now re-export `GET` so the health checks work instead of 405.

## Tests

- 10 new unit tests in `packages/integrations/src/webhooks/stripe/__tests__/paymentsSignature.test.ts` (tenant-signed, platform-signed, unknown secret, forged header, malformed payload, resolution order).
- Existing EE tenant-facade contract assertions for this file still pass.


RobertAtNineMinds:
Superseded by #3342, a clean rebuild on current main.

PR #3375, opened 2026-09-11
Fix tenant onboarding skipping support portal access for existing customers

robertisaacs:
## Summary

Tenant onboarding Step 5 unconditionally inserted the management-tenant client, so a customer whose company name already existed as a client tripped the `(tenant, client_name)` unique violation. A single wrapping catch then silently skipped contact, tag, and portal-user creation, while the welcome email went out claiming the temporary password opened the Nine Minds Support Portal — access that never existed. This was the CloudVBS / Lynda Gaiao incident.

This change makes customer tracking resolve-or-create and idempotent, isolates each provisioning stage, records real outcomes on workflow state, and gates the portal claims in the welcome email on what actually happened.

### Code changes (`ee/temporal-workflows`)

- **`activities/customer-tracking-activities.ts`**
  - `createCustomerClientActivity` now resolves an existing management-tenant client by exact, tenant-scoped `client_name`. One match → reuse untouched (bypasses `createClient` and its tax/email/notes side effects); zero → create; more than one → `AmbiguousCustomerMatchError` listing the candidate IDs (never fuzzy-link by company name). Unique-constraint violations on concurrent insert are re-read rather than failed.
  - `createCustomerContactActivity` resolves an exact-email contact scoped to the resolved client. A same-email contact under a different client raises `ContactEmailConflictError` instead of silently linking across companies (`contacts.email` is unique per tenant). Concurrent insert is re-read.
  - Returns `{ ..., reused }` so callers can distinguish reuse from creation.
- **`workflows/shared/tenant-creation-steps.ts`**
  - Step 5 is split into isolated stages (client → contact → tag → portal). A reused client/contact is success, not a swallowed error. Only an unresolvable client skips the downstream chain; tag and portal failures are recorded and non-fatal.
  - Records `clientReused`/`clientId`/`clientError`, `contactReused`/`contactId`/`contactError`, and `portalStatus`/`portalError` on `workflowState.customerTracking`.
  - Adds `AmbiguousCustomerMatchError` / `ContactEmailConflictError` to `nonRetryableErrorTypes` so an ambiguous or cross-client match fails once instead of retrying forever, and unwraps Temporal's activity-failure wrapper so state carries the real reason.
- **`db/portal-user-operations.ts`**
  - `createPortalUserInDB` looks up an existing `user_type='client'` user for the email first and returns it as `status: 'existing'` without re-hashing the password, changing roles, or reactivating it. A concurrent insert that wins the race is re-read and reused. Returns `status: 'created' | 'existing'`.
- **`activities/email-activities.ts` + `resend-welcome-email-activities.ts`**
  - `createWelcomeEmailContent` branches the intro, credentials paragraph, portal card, portal button, and "Need Help?" copy (HTML **and** plaintext) on `portalStatus`:
    - `created` → temporary credentials work for workspace **and** portal.
    - `existing` → credentials work for the workspace; an existing portal account uses its own password.
    - `failed` → portal card/button/URL omitted; copy says portal access is still being set up and offers a reply-to-this-email fallback.
    - omitted/skipped → conservative: workspace credentials only, never claims portal access.
  - `portalStatus` is optional so existing callers (resend flow) keep working and default conservatively.
- **`types/workflow-types.ts`** — `PortalProvisioningStatus`, `CustomerTrackingState`, optional `portalStatus` on `SendWelcomeEmailActivityInput`, `status` on `CreatePortalUserActivityResult`.
- **Recovery** — `src/scripts/recover-nineminds-portal-user.ts` plus `docs/runbooks/nineminds-portal-user-recovery.md`: a guarded, idempotent script that creates and links a portal user for the existing CloudVBS contact using the same shared model as production. Refuses to run without `CONFIRM_NINEMINDS_PORTAL_RECOVERY=yes`; a second run is a no-op. Production recovery is tracked separately from this code change and is not resolved until the user signs in.

### Tests

27 unit tests across `customer-tracking-activities`, `email-activities-portal-status`, `portal-user-operations`, and `tenant-creation-portal-access`. They exercise behavior (reuse, ambiguity, conflict, non-fatal portal failure, email branching), not template source strings.

## Smoke test results

**SCOPE/FIDELITY.** Real services: app on http://localhost:3254 (board dev-server, compose project `alga-psa-local-test` for Postgres/pgbouncer/Redis); the branch's real temporal worker built with `npm run build` and run from `dist` (same entrypoint as the production image); tenant creation triggered through the product's real route `POST /api/v1/tenant-management/create-tenant`; verification through the real MSP UI and real client-portal credential authentication.

**Step-downs, all disclosed:** (1) the stack's temporal container (`temporal-workflows-temporal-1`, host port 7233) is permanently stuck waiting for Cassandra, so a dedicated `temporalio/temporal:latest server start-dev` was run on 7245; (2) the welcome email's transport is Resend (a remote provider not run locally) and the repo's SMTP provider is unimplemented, so a ~30-line Resend-API simulator was run and `RESEND_BASE_URL` pointed at it — the real `ResendEmailService` code path and real rendered HTML/text bodies were exercised and captured; (3) the create-tenant UI lives in the external ninemindsreporting extension (not installed locally) and server middleware requires `x-api-key` on `/api/v1`, so an API key was minted through the real Profile > API Keys UI and used to call the route — the same auth path the extension uses; (4) the management tenant's onboarding wizard was bypassed by setting `tenant_settings.onboarding_completed` (test-data setup, unrelated to the change); (5) the portal-failure case was forced by temporarily renaming the management tenant's client-portal roles, and the ambiguous-match case by temporarily dropping `clients_tenant_client_name_unique` and inserting a duplicate client — both restored and verified restored; (6) the delivered recovery script hardcodes production IDs, so a copy with local tenant/client/contact IDs was run.

**SETUP.** Created the management tenant 'Nine Minds LLC' (`1f1946f3-75ba-42ac-8261-d193037b3852`) via the real tenant-creation workflow, logged into its MSP workspace with the credentials from its real welcome email, and built the incident precondition through the real UI: client 'CloudVBS Smoke' + contact 'Lynda Smoke <lynda@cloudvbs.smoke>' with no portal user.

**Flows and outcomes (all pass):**

1. **New-customer happy path** ('Smoke Widgets Co'): client + contact + portal user created; `portalStatus=created`; email claims workspace AND portal.
2. **Incident repro** ('CloudVBS Smoke', admin lynda@cloudvbs.smoke, client+contact pre-existing): worker logged `Reusing existing customer client` and `Reusing existing customer contact`, no duplicate-key error, portal user **created** against the pre-existing contact (`clientReused=true`, `contactReused=true`, `portalStatus=created`). Then, in the real client portal, Lynda signed in with the temporary password from the captured welcome email and reached the portal dashboard — the exact production failure, now fixed.
3. **Existing portal account preserved** ('Portal Exists Co'): portal user pre-created via the delivered recovery script; onboarding then reported `portalStatus=existing`, `hashed_password` byte-identical before/after, and the original portal password still authenticated. Email said workspace-only credentials and that a portal account already exists — no false claim.
4. **Concurrency**: two simultaneous tenant creations for the same new company name both completed — exactly one client row, both contacts and both portal users created, second run reused the client. No failure, no duplicate.
5. **Portal provisioning failure** (client-portal roles broken): tenant creation still succeeded; state carried `portalStatus=failed` with the real reason (`Client portal admin role not found for tenant`); the email dropped the portal card, button and URL entirely and said access is still being set up with a reply-to-this-email fallback; rendered HTML has no layout hole. Negative control: portal sign-in was refused and the server logged `[authenticateUser] No user found with email broken@portal.smoke` with `userType client` — the production log signature.
6. **Ambiguous customer match**: with two same-named clients, the activity failed with `AmbiguousCustomerMatchError`, Temporal recorded `RETRY_STATE_NON_RETRYABLE_FAILURE` (one attempt, no retry storm), no contact was created under either candidate, the workflow completed, and the email made no portal claim. Candidate IDs are in workflow state.
7. **Tenant-isolation guard**: onboarding a new company with an admin email already used by a contact under a different client raised `ContactEmailConflictError`; the existing contact stayed attached to its original client, no portal user was created, no cross-company link, tenant creation still completed.
8. **Nearby-regression check**: resend-welcome-email through its real route sent a conservative email (no portal section, workspace-only credential claim), matching the documented decision.
9. **Recovery procedure**: the delivered script refused to run without `CONFIRM_NINEMINDS_PORTAL_RECOVERY=yes`; first run created the portal user and printed a one-time password; second run printed `Portal user already exists ... No changes made` with the password hash unchanged.
10. **Unit tests**: all 27 pass.

**Could not test / observations (none release-blocking):** (a) a full workflow re-run with the SAME admin email never reaches customer tracking — it fails earlier at `createAdminUser` (`internal user email must be unique across all tenants`) and retries indefinitely because that error is not classified non-retryable, so the API returns a 2-minute "still running"; pre-existing, untouched by this branch, covered instead via the reuse paths, the concurrent race, and recovery-script re-run. (b) `AmbiguousCustomerMatchError` is unreachable while `clients_tenant_client_name_unique` exists (the constraint was dropped to reach it); sound defence-in-depth for schemas without it. (c) When the client step fails, `portalStatus` stays `skipped`, so the email uses the neutral reply-to-this-email copy rather than the "still being set up" copy; the real reason is in workflow state either way. (d) The resend flow fails in `logAuditEvent` when `MASTER_BILLING_TENANT_ID` is unset in the worker env — a pre-existing worker env requirement, not this change. (e) The insert-then-conflict branch of client creation was not reached in the live race (the pre-check won both times); unit-tested, and the observable guarantee (single client, no failure) held.

**Cleanup.** Temporary temporal container removed, worker and simulators stopped, `server/.env.local` restored from backup and the dev-server restarted (200 on `/auth/msp/signin`), dropped constraint restored and duplicate rows deleted, portal roles renamed back, worktree git status clean. Evidence directory: `/tmp/alga-smoke-evidence/tenant-onboarding-portal-20260911-110135`.

## CI unblock (pre-existing, unrelated)

The first CI run failed one test out of 15,208: `packages/billing/tests/clientContractEffectiveRenewalSettings.test.ts > respects evergreen-specific notice period overrides when tenant defaults are disabled` (`expected '2027-08-11' to be '2026-08-11'`). This is a pre-existing time bomb on `main`, not caused by the onboarding change (`packages/billing` and `shared/billingClients` are untouched by this branch).

Root cause: `normalizeClientContract` resolves the evergreen review anchor against the real clock (it accepts no `now`), while the test hardcoded `now = '2026-01-01'` for its expected dates. Once wall-clock time passed the 2026-09-10 anniversary, the next anchor rolled to 2027. The fix pins `Date` to `2026-01-01` for just that test (`vi.useFakeTimers({ toFake: ['Date'], now })` inside `try/finally`), preserving the tenant-default vs contract-override assertion. Test-only; no production behavior change. Verified locally: all 19 tests in the file pass.

The four Temporal test files this branch added were collected by no runner, so the Repository test inventory reported `No runner collects test` and Production regression readiness failed. They are now registered in `scripts/lib/test-discovery.mjs` (lane candidates) and each lane's vitest include: the three activity/DB-double tests (`customer-tracking-activities`, `email-activities-portal-status`, `portal-user-operations`) in `temporal-readiness`, and the workflow test that spins an ephemeral Temporal test server (`tenant-creation-portal-access`) in `temporal-engine`. Verified locally with the exact CI commands: temporal-readiness 48 files / 298 tests pass, temporal-engine 13 files / 67 tests pass.

## Scope

This PR covers the onboarding bug and the affected-account recovery tooling. Production recovery for Lynda is tracked separately and is not considered resolved until her portal access is verified. No schema changes. The CI-unblock commit is a test-only determinism fix for a pre-existing date-bomb and can be dropped independently of the product change.


robertisaacs:
## CI diagnosis: `browser / fresh-install-e2e` red is a partial-re-run artifact, not a code defect

The two red checks (`browser / fresh-install-e2e` and `Production regression readiness`) have a single deterministic cause, and it is not this branch.

**What happens.** `scripts/lib/browser-artifact-manifest.mjs` pins browser evidence to a run *attempt*: `identity()` includes `runAttempt`, and `validateBrowserArtifactManifest` rejects a manifest whose `runAttempt` differs from the gate's `GITHUB_RUN_ATTEMPT`. That rejection sets `artifactFailure` in `browserTestMetrics`, yielding `status: 'incomplete'`, which the readiness check reports as the opaque `provider readiness: browser-execution-not-passed`.

Re-running *only failed jobs* bumps `GITHUB_RUN_ATTEMPT` for the gate while leaving the already-green producer jobs' artifacts stamped with the older attempt. The gate can then never pass.

**Evidence** (artifacts downloaded from run 34630360431):

| attempt | community manifest | enterprise manifest | gate runs at | gate failure |
|---|---|---|---|---|
| 1 | 1 | 1 | 1 | real: `production-browser: failure` |
| 2 | 2 | 1 | 2 | enterprise only |
| 3 | 2 | 1 | 3 | community **and** enterprise |

The attempt-2 result is the decisive one: it failed on enterprise alone, precisely because community's manifest had been regenerated at attempt 2 while enterprise's was still attempt 1. The failure tracks manifest-vs-attempt skew exactly, so it is deterministic rather than flaky.

**The original failure was different and real:** at attempt 1, `integration / Infrastructure shard 1` and `browser / Production browser (community)` failed. Both passed on re-run, so they were infrastructure flakes. Every failure since has been the identity-skew artifact above.

**The five "unrelated failing workflows" are one failure.** `supported-upgrade-execution`, `supported-citus-upgrade-execution`, `microsoft-callback-execution`, `teams-development-execution` and `fresh-install-execution-gate` all declare `job: 'browser'` in `scripts/lib/production-readiness.mjs`. When the `browser` job fails, each reports `Required workflow: failure` independently. They are one root cause reported five times, not five problems.

**Remedy applied:** re-ran **all** jobs (attempt 4) so producers and gate share one attempt. Same-name artifacts coexist across attempts (three `fresh-install-execution-gate` artifacts already do), so no upload conflict.

**Operational note for whoever handles this next:** if this run goes red on a flake, re-run **all** jobs again. "Re-run failed jobs" cannot fix this gate once a producer job has emitted artifacts under an earlier attempt — it is guaranteed to stay red.

The attempt-pinned gate being unfixable by partial re-run is a latent CI fragility worth addressing on its own; it is deliberately out of scope for this card and untouched here.

robertisaacs:
CI fix-in-place (head now `8ebac8a7b1`). The prior red run (SHA `e78c92c1ac`) had 6 failures, all traced to two causes now resolved:

1. **workspace-runtime → Workspace gate, Repository test inventory, Production regression readiness.** Root cause was `docker: pull access denied for minio/minio` — the old workflow pulled MinIO from Docker Hub. `origin/main` fixed this in `5220b3094f (fix(ci): pull MinIO images from quay.io)`; merged `origin/main` into the branch to pick it up. The inventory 'No runner collects test' entries (s3-bundle-store, WorkflowRuntimeV2TemporalWorker — both pre-existing on main) were a cascade from the missing workspace-runtime `evidence.json` when that job crashed; they resolve once the lane runs.
2. **integration shard 3 → Integration execution complete.** `appointmentAvailability.integration.test.ts > partial overlap detection` failed on `duplicate key value violates unique constraint service_types_tenant_order_number_unique`. Pre-existing flake: fixtures used `Math.floor(Math.random()*1000000)` for `order_number`, which collided across the serial suite's shared tenant. Fixed deterministically with a max(order_number)+1 allocation (`8ebac8a7b1`).

Both are unrelated to the onboarding product change and separable; disclosed here for review. No product-code changes in this update.

